Issue #5, opened 2022-03-20
Language field - no ption All, default language selected


szepty-ziemi:
FEATURE
Language selector field in category creation form
AND
Language selector field in category creation form

GIVEN
Having more than one content languages
AND
User is in Category Manager
OR
User is in Image Manager
WHEN
Clicking New button
AND
Viewing language selector field
THEN
Default language is selected
BUT
Option All should be selected by default

Prerequisites
![obraz](https://user-images.githubusercontent.com/60470150/159170777-1ea2a605-5a2f-43f9-8af8-4aa4827f58fe.png)

![obraz](https://user-images.githubusercontent.com/60470150/159170717-6f7a14f2-209a-4648-a9b1-79e8e59f475d.png)
![obraz](https://user-images.githubusercontent.com/60470150/159170726-2a27b3dd-7f93-47fe-8d49-c8a71f1b4951.png)

![obraz](https://user-images.githubusercontent.com/60470150/159170740-c6205963-888c-413c-9b3d-30bd98b07a11.png)

We should replicate how language selector field works in com_content:
![obraz](https://user-images.githubusercontent.com/60470150/159170842-8fbea1fe-d524-45d3-80da-70400acf0910.png)

Also in com_content language field appears after language plugin _System - Language Filter_ is enabled.

MrMusic:
@szepty-ziemi This Issue should be fixed now with PR https://github.com/JoomGalleryfriends/JG4-dev/pull/7. Can we close ths Issue?

szepty-ziemi:
@MrMusic please, wait till this PR will be merged https://github.com/JoomGalleryfriends/JG4-dev/pull/10 
With current merges with main branch I still have issues with installation so any tests are not reliable.
So right after PR merge I mentioned I will test this issue and close it accordingly.

szepty-ziemi:
I confirm, issue is fixed after merge with PR #7 and #10 .
Tested on latest main branch pull.

PR #7, opened 2022-03-26
fix language field behavior

Elfangor93:
Fixes the bug reported in issue #5


MrMusic:
Thank you, 
In principle, it works for me.
One small feature is still missing: In Joomla, the language field in the list views is hidden if multilanguage is not activated.
Can we do it in that way too?
See screenshot from com_categories
JG4-with-multilanguage
JG4-without-multilanguage

MrMusic:

PR #45, opened 2022-10-10
Add Hide Table Columns for categories and images

MrMusic:
Since joomla 4.2 there is a nice feature to hide any column in the table views: https://github.com/joomla/joomla-cms/pull/36591

@szepty-ziemi had already suggested the implementation at the last online meeting.
Surprisingly, the implementation in the JoomGallery was very simple.
Used for categories and images view because in tags and configuration view there are only a few columns.

### DIFF
--- a/administrator/com_joomgallery/tmpl/categories/default.php
+++ b/administrator/com_joomgallery/tmpl/categories/default.php
@@ -29,6 +29,7 @@ $wa = $this->document->getWebAssetManager();
 $wa->useStyle('com_joomgallery.admin')
    ->useScript('com_joomgallery.admin')
    ->useScript('com_joomgallery.catBtns')
+   ->useScript('table.columns')
    ->useScript('multiselect');
 
 $user      = Factory::getUser();
--- a/administrator/com_joomgallery/tmpl/categories/default.php
+++ b/administrator/com_joomgallery/tmpl/categories/default.php
@@ -69,7 +69,7 @@ if($saveOrder && !empty($this->items))
                   <?php echo HTMLHelper::_('searchtools.sort', '', 'a.lft', $listDirn, $listOrder, null, 'asc', 'JGRID_HEADING_ORDERING', 'icon-sort'); ?>
                 </th>
                 <th scope="col" class="w-1 text-center">
-                  <?php // Spaceholder for thumbnail image ?>
+                  <?php echo Text::_('COM_JOOMGALLERY_IMAGE') ?>
                 </th>
                 <th scope="col" class="w-1 text-center">
                   <?php echo HTMLHelper::_('searchtools.sort', 'JSTATUS', 'a.state', $listDirn, $listOrder); ?>
--- a/administrator/com_joomgallery/tmpl/categories/default.php
+++ b/administrator/com_joomgallery/tmpl/categories/default.php
@@ -46,17 +46,20 @@ if($saveOrder && !empty($this->items))
 }
 ?>
 
-<form action="<?php echo Route::_('index.php?option=com_joomgallery&view=categories'); ?>" method="post" name="adminForm" id="adminForm">
-  <div class="row">
-    <div class="col-md-12">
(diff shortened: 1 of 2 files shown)


szepty-ziemi:
ENV: https://prnt.sc/uVfTvuEhFb4v

1. Dropdown column list causes to scroll bar to appear on the image/categories list
- go to image / category manager
- click column button

RESULT
- dropdown cause scrollbar to appear

EXPECTED RESULT
- dropdown should appear and be visible, no scrollbar displayed

obraz

Issue do not appear if there is enough items on the list, and list is longer than dropdown.

obraz

szepty-ziemi:
ENV: https://prnt.sc/uVfTvuEhFb4v

2. There is unnecessary checkbox for list item checkbox column. This is spared and not displayed in Joomla core components.

- just click on column selector to display dropdown

obraz

Issue apply to both Image and Category manager.

Here is how it looks in Joomla core:

obraz

szepty-ziemi:
While testing on Windows, I encountered issue right after installation. 
- Install JG package
- go to Image Manager or Category manager

Only those two places are affected. All other e.g Configuration Manager are ok.

obraz

ENV: https://prnt.sc/5YN3u5ANetf7

MrMusic:
1. >Dropdown column list causes to scroll bar to appear on the image/categories list

This effect is due to the fact that other classes are currently used in JoomGallery than in the corresponding Joomla views. I don't know if this is intentional?
Now i changed the classes for the images and categories view so that it is the same as in the Joomla views.
Maybe this will have any negative effect?

MrMusic:
> Only those two places are affected. All other e.g Configuration Manager are ok.
> 
> obraz

The only possible reason for me: You are not using Joomla **4.2**.x but an older version.

szepty-ziemi:
@MrMusic BINGO! That was the source of issue. Upgraded to latest version and issue was gone.

szepty-ziemi:
I confirm that latest commit was tested on both Linux and Windows. Everything works and looks good.

administrator/com_joomgallery/tmpl/categories/default.php
```diff
@@ -45,20 +46,17 @@
...
-				<?php echo LayoutHelper::render('joomla.searchtools.default', array('view' => $this)); ?>
-				<div class="clearfix"></div>
-        <div class="table-responsive">
-          <table class="table table-striped" id="categoryList">
+<form action="<?php echo Route::_('index.php?option=com_joomgallery&view=categories'); ?>" method="post" name="adminForm" id="adminForm">
+  <div class="row">
+    <div class="col-md-12">
+      <div id="j-main-container" class="j-main-container">
+        <?php echo LayoutHelper::render('joomla.searchtools.default', array('view' => $this)); ?>
+          <table class="table" id="categoryList">
```
Elfangor93, line 54: Whats the reason for removing the "table-striped" class?

Elfangor93:
>Now i changed the classes for the images and categories view so that it is the same as in the Joomla views.
Maybe this will have any negative effect?

@MrMusic Since you are changing only the template files of images and category manager, I can not see any negative side effects except for the fact that it is displayed differently as the first implementation was suggested.

administrator/com_joomgallery/tmpl/categories/default.php
MrMusic: I have changed the classes so that it is the same as in Joomla.
The missing scrollbars are actually a disadvantage here.
Now i have reversed it.

MrMusic:
>

> ENV: https://prnt.sc/uVfTvuEhFb4v
> 
> >    Dropdown column list causes to scroll bar to appear on the image/categories list
> 
> >    go to image / category manager
> >    click column button
> 
> RESULT
> 
>     dropdown cause scrollbar to appear
> 
> EXPECTED RESULT
> 
>     dropdown should appear and be visible, no scrollbar displayed

Sorry, the fix also prevented the horizontal scrollbars from appearing. so i had to reverse it.

Issue #11, opened 2022-04-03
Action buttons should be grupped under "Actions" button dropdown (Image, Category, Tags and Configuration Manager)


szepty-ziemi:
Test Env:

PHP Built On: Linux s47.smarthost.pl 4.18.0-348.12.2.lve.el8.x86_64 #1 SMP Mon Jan 24 07:29:58 EST 2022 x86_64
Database Version: 10.3.34-MariaDB
PHP Version: 7.4.28
Web Server: Apache
Joomla! Version: Joomla! 4.1.2 Stable [ Kuamini ] 30-March-2022 16:00 GMT
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:98.0) Gecko/20100101 Firefox/98.0 

Current view in JoomGallery:
![obraz](https://user-images.githubusercontent.com/60470150/161427849-f4156534-571a-4fa1-9116-931606b9ba05.png)

View in Joomla categories:
![obraz](https://user-images.githubusercontent.com/60470150/161427861-99a974ca-3885-49cc-9ee0-dd0a574dbe3e.png)

I guess this discrepancy will be part of functional buttons revamp in future, so this ticket could be expanded with full discrepancies description.

szepty-ziemi:
Issue is also valid for Configuration Manager:

![obraz](https://user-images.githubusercontent.com/60470150/174073567-0b9cea99-7005-47a8-89e7-cd77fe44ea96.png)

Elfangor93:
This issue should be fixed with PR#31.
@szepty-ziemi please confirm and close issue if so.

szepty-ziemi:
Current view (main branch, installed 07.10.2022) looks differently than originally reported. Nevertheless it still looks different than category manager in Joomla. My suggestion is to merge:

- Publish
- Unpublish
- Duplicate
- Delete

Into one, under Action button:

![obraz](https://user-images.githubusercontent.com/60470150/194589192-6148bca3-f1f0-4c14-a7ee-4f9a9913ca3b.png)

Just like in Joomla article categories:

![obraz](https://user-images.githubusercontent.com/60470150/194589277-4e9bb9df-1a6a-42c8-9693-dcfc2864f714.png)

The same apply to Image Manager:

![obraz](https://user-images.githubusercontent.com/60470150/194589590-9ba684c3-504c-4597-8b2d-807841f91b11.png)

And in Joomla articles:

[...]

PR #46, opened 2022-10-14
Reorganise toolbar buttons

MrMusic:
Possible fix for https://github.com/JoomGalleryfriends/JG4-dev/issues/11
The toolbar buttons now have the same order and identical names in all managers:

- New
- Batch
  - Duplicate
  - ...
- State
  - Publish
  - Unpublish
- Delete

@Elfangor93 Do we need a 'state' (Publish) column for configuration(s) or can this column deleted?

### DIFF
--- a/administrator/com_joomgallery/src/View/Categories/HtmlView.php
+++ b/administrator/com_joomgallery/src/View/Categories/HtmlView.php
@@ -71,33 +71,74 @@ class HtmlView extends JoomGalleryView
 		parent::display($tpl);
 	}
 
-	/**
-	 * Add the page title and toolbar.
-	 *
-	 * @return  void
-	 *
-	 * @since   4.0.0
-	 */
-	protected function addToolbar()
-	{
-		$state = $this->get('State');
-		$canDo = JoomHelper::getActions('category');
+  /**
+   * Add the page title and toolbar.
+   *
+   * @return  void
+   *
+   * @since   4.0.0
+   */
+  protected function addToolbar()
+  {
+    $state = $this->get('State');
+    $canDo = JoomHelper::getActions('category');
 
-		ToolbarHelper::title(Text::_('COM_JOOMGALLERY_CATEGORY_MANAGER'), "folder-open");
+    ToolbarHelper::title(Text::_('COM_JOOMGALLERY_CATEGORY_MANAGER'), "folder-open");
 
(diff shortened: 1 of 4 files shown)


Elfangor93:
@MrMusic I would like to have a publish column in the configuration manager.

MrMusic:
>@MrMusic I would like to have a publish column in the configuration manager.

Ok, i changed the toolbar in the configuration manager so that it is identical to the other managers.
Because the functionality in the configuration is currently changed, the missing functions should be built in there: https://github.com/JoomGalleryfriends/JG4-dev/pull/47

szepty-ziemi:
This looks very good now. One question: could we have different icons for button groups, that currently have three dots?

I think it will be good to distinguish them with different icons. Joomla 4 is using FontAwesome 5. Maybe we could use some of their fancy icons? :) Here are my suggestions:

**Batch**

https://fontawesome.com/icons/reply-all?f=classic&s=solid
https://fontawesome.com/icons/recycle?f=classic&s=solid
https://fontawesome.com/icons/shuffle?f=classic&s=solid

**Status**

https://fontawesome.com/icons/circle-dot?f=classic&s=regular
https://fontawesome.com/icons/toggle-on?f=classic&s=solid
https://fontawesome.com/icons/circle-check?f=classic&s=regular

**Image Manipulation**

https://fontawesome.com/icons/sliders?s=solid&f=classic
https://fontawesome.com/v5/icons/images?s=regular&f=classic

MrMusic:
@szepty-ziemi Thank you for your suggestions.

**Batch**
I think the icons always suggest a certain function to the user. 
However, "Batch" can mean various functions.
The icon should therefore be kept quite general or have a reference to the respective manager.

My suggestion would be to use the icon from the header for the respective manager. It is implemented like this for now.
But maybe we should also leave the icon with the three dots to be compliant with Joomla?

**Status**
For the state group, I have now taken the publish symbol with circle that you suggested:
https://fontawesome.com/icons/circle-check?f=classic&s=regular

**Image Manipulation**
For the image manipulation group in image manager I have now taken this one that you suggested:
https://fontawesome.com/v5/icons/images?s=regular&f=classic

szepty-ziemi:
@MrMusic I'm good witch your choice 😸

Elfangor93:
I tested this PR successfully. Thank you...

PR #49, opened 2022-11-12
New field: JG-Image selection

Elfangor93:

It adds a new field to JoomGallery which allows to select a JoomGallery image using a modal box.
Clicking the button opens a modal box where you can search for the image. A click on the image name closes the modal and sets the image as new value in the field. A thumbnail shows the currently selected image.
grafik

The field is used in the category edit form to select a thumbnail image of the category.

Go to category manager, edit a category and select a thumbnail for the category. Use filter and sorting options in the provided modal box to search for the correct image in your gallery.


eumel1602:
Unfortunately it doesn't work for me. It's just an input field. I can't click anything
(J 4.2.5. PHP 7.4 Windows)

Elfangor93:
Ahh shit. The same problem as with the image processor dropdown field in the config manager.
I code on the case insensitive Windows and you are testing on the case sensitive Linux.
I will fix it as soon as I am back home...

Elfangor93:
@eumel1602 please try again. Should be fixed now...

eumel1602:
If I want to select a preview image and sort by: "Select status" by "no main entry" then the images that are a "main entry" are also displayed. In addition, I sort in "Select traffic jams" by: "Hidden" are also the pictures that are "Published" for selection (see picture)

I couldn't find any other errors... (J 4.2.5. / PHP 7.4 / Windows / Online test environment

[bild1](https://user-images.githubusercontent.com/96334985/201524005-a1c1aee4-16e9-459a-a87c-8fe6636fd88c.jpg)

Elfangor93:
> I sort in "Select traffic jams" by: "Hidden"

Sorry, I have no clue what you mean by that.

MrMusic:
It would be nice if the images were filtered according to the 'current' category when the pop-up is called. In most cases, the user wants to select a image from the current category.

AlexanderSupp:
It works fine after a SAVE. But when I select an image and close the pop-up there is no refresh of the screen. Only after SAVE I see the result. This is confusing.

MrMusic:
@AlexanderSupp yes, you are right.
There is a (small) problem if 'no image' was saved before. Then, after selecting an image, only the title is updated but not the thumbnail.
If an image/thumbnail was already saved, it works without any problems.

Elfangor93:
>There is a (small) problem if 'no image' was saved before.

Haha. It looks may look like a small problem, but it was one of the trickiest to solve so far...

MrMusic:
>Haha. It may look like a small problem, but it was one of the trickiest to solve so far...

Thanks for fixing this.

Unfortunately, since the last update, the Category filter is not respected when the pop-up is called.

### Steps to reproduce the issue

Create new Categories and add images to these Categories.
Open a Category and click 'Select a image'.

### Expected result

Only images from the current Category should be displayed.

### Actual result

All images are shown.
The Category filter is only taken into account after an 'refresh'.

wrong_category_filter

Elfangor93:
@MrMusic are your found issues resolved. Merge?

MrMusic:
In Image Manager is a error now. 
See call stack:
image-manager-error

MrMusic:

PR #50, opened 2022-11-13
New field: JG-Category selection

Elfangor93:

It adds a new field to JoomGallery which allows to select a JoomGallery category using a modal box.
Clicking the button opens a modal box where you can search for the category. A click on the category name closes the modal and sets the category as new value in the field.
grafik

The field is used in the category edit form to select a parent category.

Go to category manager, edit a category and select a parent category. Use filter and sorting options in the provided modal box to search for the correct category in your gallery.

### DIFF
--- a/administrator/com_joomgallery/forms/category.xml
+++ b/administrator/com_joomgallery/forms/category.xml
@@ -48,8 +48,8 @@
            description="JFIELD_ALIAS_DESC" />
 
     <field name="parent_id"
-           type="categoryedit"
-           default="0"
+           type="jgcategory"
+           default="1"
            label="JGLOBAL_SHOW_PARENT_CATEGORY_LABEL" />
 
     <field name="published"
--- a/administrator/com_joomgallery/forms/category.xml
+++ b/administrator/com_joomgallery/forms/category.xml
@@ -49,6 +49,7 @@
 
     <field name="parent_id"
            type="jgcategory"
+           children="false"
            default="1"
            label="JGLOBAL_SHOW_PARENT_CATEGORY_LABEL" />
 
--- a/administrator/com_joomgallery/forms/filter_categories.xml
+++ b/administrator/com_joomgallery/forms/filter_categories.xml
@@ -18,7 +18,7 @@
 
         <field
               name="category"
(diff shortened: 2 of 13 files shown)


Elfangor93:
Info for the testers:
grafik
In the modal box when you want to filter for a specific user, there is no dropdown to select available users. This is due to the fact that this dropdown field is added in another PR.
This is not a bug this is intended to be like that. As soon as this PR is merged the dropdown will be available.

eumel1602:
I couldn't find any errors
  (J 4.2.5. / PHP 7.4 / Windows / Online test environment

eumel1602:
It was not possible for me to sort the categories according to the number of pictures in the categories themselves!?

Elfangor93:
Yes you are correct that is not possible. And I also will not implement that at that statge. We have to focus on the most important functions.

MrMusic:
In category edit form if no parent category is selected, the term 'ROOT' should not appear. This may confuse some users.
Instead, a text like 'No parent category' or like in Joomla '- No parent -' should appear.

The same text should also appear in the (no category) button in the Popup.

szepty-ziemi:
When trying to assign category to itself, we should have a bit more meaningful error message:

obraz

Eg. "You can't assign category to itself" or "You can't make child category of itself"

Elfangor93:
> The same text should also appear in the (no category) button in the Popup.

No. The view in the popup will not only be used to select parent categories. Therefore I prefere having a button text which can be used everywhere in JoomGallery not only to select a partent category.

Elfangor93:
> When trying to assign category to itself, we should have a bit more meaningful error message:

I dont want to override the error message since that would mean to override core classes which would blow up our code base unnecessarily. Instead I am planning to add a possibility to the field to exculde child categories in the list view (popup). That way users will not be able to select child categories and this error should not be raised anymore.

Elfangor93:
@szepty-ziemi Do you like the solution to the error message issue?

szepty-ziemi:
> @szepty-ziemi Do you like the solution to the error message issue?

That's a perfect solution :) Thank you.

obraz

PR #51, opened 2022-11-19
New field: JG-Tag selection

Elfangor93:

It adds a new field to JoomGallery which allows to select JoomGallery tags.
The field behaves like the tags selection field of the Joomla Core. Meaning, you get a dropdown field with search possibilities and if you have the permission to add new tags.
grafik

The field is used in the image edit form to select tags.

Go to image manager, edit an image and assign tags to the image.
Remove already assigned tags. Use the possibility to add new tags.
Generally test, if the field behaves the same way as the tags field of Joomla Core.


AlexanderSupp:
Edit image, Tags field is in Tab "Details".
From my point of view it is better in Tab "Images" where I can see the image here.
Sort order of tags alphabetical would be helpfully.

The function works.
I can simply add new tags in image edit and the sync with Tags Manager is correct.

Btw.
Is it possible for me to test all open pull requests in one system?
How can I install all functions?

Elfangor93:
>Sort order of tags alphabetical would be helpfully

Currently the tags are sorted the same way as the tags from Joomla core. Meaning: The most used tags at the top.

> Is it possible for me to test all open pull requests in one system?

You can test one pull request after another on the same system. Having all new functions at ones is only possible after the pull requests are merged into main.

Elfangor93:
@eumel1602 If I am correct that is also the case in the core tags, right?

I guess you mean unpublished not hidden 😉.

eumel1602:
Yes that's right. I can answer both questions with yes.

Elfangor93:
I think we should stay as close as possible to Joomla core functionality.

And anyway I can see usecases where you want to add unpublished tags to your images...

eumel1602:
It wasn't a criticism. It was a statement. if that's the way it should be why not.

MrMusic:
Works with images that are already stored (edit image).
But the Tags are not stored when the image is 'new'.

## Steps to reproduce the issue
Click 'New' in Image Manager
Type in title, description...
Select image
Add Tags
Click Save & Close

## Expected result
The Tags should be storedin image

## Actual result
The Tags are not stored

## System information
Joomla 4.2.5
PHP 8.0

szepty-ziemi:
Everything seems to work well, except issue reported by @MrMusic I noted some differences between core Tag feature and Joomla Gallery:

obraz

We don't need Type as only images have tags (or we plan this for categories as well?)

We don't need Select Max Levels as we have only one level of tags (or we plan to provide assigning parents for tags?)

As for statuses I wonder do we need Trash and Archive features.

What about feature for hiding columns?

szepty-ziemi:
Found a bug:

1. Go t image
2. In tag field type <test>
3. Click Enter

obraz

Result:

Empty tag will be created in tag field

obraz

Here we can see two attempts of such a tag creation:

obraz

Expected result:

1. Tag "<test>" should be saved OR
2. < should be stripped and just "test" should be saved

ENV: https://prnt.sc/1SCzlH0TMaQZ

szepty-ziemi:
Tag use counter shows wrong value:

1. Create two images and three different tags
2. Assign all tags to first and second image
3. Delete one of the images
4. Check Tags use counter in Tags Manager:

obraz

obraz

obraz

obraz

ENV: https://prnt.sc/1SCzlH0TMaQZ

Elfangor93:
#### Comment to the issue "Tag counter shows wrong value"
The problem here was, that the DB table #__joomgallery_tags_ref was not updated correctly when deleteing images. So please clean up #__joomgallery_tags_ref before further testing. Otherwise it could lead to strange behavior.

How to clean up #__joomgallery_tags_ref:
- Delete all tags references in your images by emptying tag field in image edit form
- Delete all records in records in #__joomgallery_tags_ref table

Elfangor93:
No sanitation of titles is a general problem. I will not solve it in this PR but I opened an Issue instead: #56

MrMusic:
>The problem here was, that the DB table #__joomgallery_tags_ref was not updated correctly when deleteing images

The DB table #__joomgallery_tags_ref is now correctly updated when an image is deleted.
When deleting a Tag, the assignment in table #__joomgallery_tags_ref should also be deleted.

MrMusic:
and it would be nice if you could 'filter' by Tags in the Image Manager - like in Joomla.

Elfangor93:
>and it would be nice if you could 'filter' by Tags in the Image Manager - like in Joomla.

Added an issue for that #57

MrMusic:

PR #59, opened 2022-12-27
Checkout/Checkin functionality

Elfangor93:

Adds the checkout and checkin functionality to image, category, config and tags manager as you know it from com_content. If a user is editing a record, the record gets a lock in the list view. This record is now locked for anybody else to edit until it is unlocked again.

- Edit a record by opening a second tab (image, category, config set or tag)
- Close tab without closing the edit form with the proper button
- Refresh the list view. The edited record should have now a lock in front of the title.
- If you have the correct permission you can unlock the record byclicking on the lock.


AlexanderSupp:
Very important for professional systems. But for an image gallery in Joomla? However, I will test it.

eumel1602:
I wanted to test it, every time I try to create an image or a category I get the following error:
Save failed with the following error: Field 'checked_out_time' doesn't have a default value

is it because of my installation or because of the JG4???

MrMusic:
>"Save failed with the following error: Field 'checked_out_time' doesn't have a default value"

>is it because of my installation or because of the JG4-PR Checkout/checkin???

Same problem with me. Error is with all tables: images, categories, tags, configs.

eumel1602:
(Now the installation seems to be ok.)
I've tried the PR with many different tabs and settings. I also installed and tested the sample data from Mr Music. (50 to 100 tests)
Except for very few cases, I was always able to see the lock I wanted or it was gone again at the right time.
Very rarely it wasn't true, the post wasn't blocked, I could never reproduce it.
From my point of view it does what it should in the 4 managers ...

MrMusic:

PR #60, opened 2023-01-01
Improved messaging trait

Elfangor93:
## What is this trait for

The messaging trait adds a session based message handling to the component class (Joomgallery\Component\Joomgallery\Administrator\Extension\JoomgalleryComponent).
It allows to collect warnings, debug information and error and output them as one message even over multiple reloads of the page. This is needed for operations with a long runtime needing multiple site refreshes to complete.
In JG3 there was the debugoutput which acted in a similar way. This behaviour will be implemented with that in a more robust and expandable way.

## How to test this PR
Create new images and categories. The behavior and the printed messages should not change from before this PR.


AlexanderSupp:
Sounds good. Will watch that.

eumel1602:
I have created different images and categories. Produced one or the other error himself. Everything was properly displayed to me in the message.
I can't see what exactly needs to be tested.
The expected result: "The behavior and the printed messages should not change from before this PR." I don't understand.
Maybe a more detailed explanation of what could be wrong or what to look out for...

MrMusic:
I think you can't do more than check whether other or wrong messages appear.
In my tests I could not see any problem.
I think it should be merged.

eumel1602:
> I think you can't do more than check whether other or wrong messages appear. In my tests I could not see any problem. I think it should be merged.

ok, then I can't find anything wrong

Issue #43, opened 2022-10-08
Category - +Neu, Input Name of Category, +SPEICHERN & NEU - results in "Speichern fehlgeschlagen. Fehler: Field 'modified_time' doesn't have a default value"


AlexanderSupp:
Category - +NEU
input categorie name
+SPEICHERN & NEU
results in "Field 'modified_time' doesn't have a default value"
If you select an existing category entry and hit +SPEICHERN & NEU no such error happens.
![CategorySpeichernUndNeu-Screenshot 2022-10-08 141435](https://user-images.githubusercontent.com/110494397/194707449-61d8f9f8-2cd6-4c52-85da-02d729aa01f4.jpg)

PR #71, opened 2023-01-22
Group action buttons in edit forms

MrMusic:
In edit forms Joomla! groups 'Save & Close', 'Save & New' and 'Save as Copy' as a dropdown.
group-action-buttons

This pr do this in the same way.

### DIFF
--- a/administrator/com_joomgallery/src/View/Category/HtmlView.php
+++ b/administrator/com_joomgallery/src/View/Category/HtmlView.php
@@ -70,6 +70,8 @@ class HtmlView extends JoomGalleryView
 	{
 		Factory::getApplication()->input->set('hidemainmenu', true);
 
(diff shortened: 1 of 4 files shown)


eumel1602:
When testing, I noticed that the "Save & New" button does not save the images and brings an empty field but brings an error message.
This happens with the pictures, categories and also tags
bildschirmabgriff32

MrMusic:
Thanks for testing @eumel1602 
The cause of this error does not lie in this PR but has been known for some time, see https://github.com/JoomGalleryfriends/JG4-dev/issues/43
As a workaround, a new item must first be saved once.

eumel1602:
OK. Then I couldn't find any more error.
Tested with J4.2.5 / PHP 7.4

Elfangor93:
Testes successfully ✅. Thank you very much.

PR #76, opened 2023-02-06
Fix error when memory_limit is set to -1

MrMusic:
The function checkMemory checks the value of memory_limit which is set in php.ini. Some hosters have set a value of -1 here. This 'unusual' value causes the check to fail.
This is a redo from https://github.com/JoomGalleryfriends/JoomGallery/pull/123


Elfangor93:
Code changes look good to me.

Elfangor93:
I tested this successfully. Thank you very much.

PR #79, opened 2023-03-05
Tus upload service

Elfangor93:
## What is this service for
The TUS server service will be used to upload multiple files in a Drag'N'Drop form input field.
The core of this service is an upload server implementing the [TUS resumable upload protocol](https://tus.io/protocols/resumable-upload.html). The TUS server provides a backend to receive files from any upload client implementing the TUS protocol.

### Backend implementataion
The TUS server in the backend was created using https://github.com/Orajo/zf2-tus-server. The code was modified to use [Joomla\CMS\Application\CMSApplication](https://api.joomla.org/cms-3/classes/Joomla.CMS.Application.CMSApplication.html) instead of the laminas framework to create the http responses.

### Frontend implementation
The uploading form in the fronted contains the upload form field [Uppy](https://uppy.io/examples/dashboard/). This is the official frontend upload client from the developer of the TUS protocol.

## How does it work
Upload requests are sent from the frontend client via XHR request to the URL `/administrator/index.php?option=com_joomgallery&task=images.tusupload`. This will start the TUS server handling the incoming requests and saves the approaching data to the Joomla temp folder. The uploaded files are saved with the Upload-ID (UUID) as filename.

## How to use this service
The service gets loaded automatically when you send a request with the task `task=images.tusupload` to the JoomGallery component in the Joomla backend.
But you can load the service anywhere in Joomla by doing

```
$this->component->createTusServer();  // instantiate service
$server = $this->component->getTusServer();  // load the service into your scope
$server->process(true);  // run tus server
```

When instantiating the service you could add a parameter to the createTusServer() method specifying the folder where uploaded files should be stored. Default upload folder is the Joomla temporary (temp) forlder.

## How to test the service

Navigate to Components-JoomGallery-Images in the Joomla backend.
grafik
The "New"-Button now has a ned sub-button called "Multiple New". This button brings you to the upload from with an uppy form field. Start uploading files using this uppy form field.
After upload a button with debug information is added to the thumb of theuploaded file:
grafik
Read out the Upload-ID (UUID) in the Debug information and check the following:

- A file is created on the server in the upload folder (Joomla temp folder) with the UUID as filename.
- A second file is created called <UUID>.info containing different informations of the uploaded file.
- Rename the created file (with UUID as filename) and add the file extension to the filename.
- Open the renamed file and check its content.

### Normal file upload
You can upload any files up to a size of 250MB. Try uploading different file types. The files gets uploaded into the temp folder of Joomla. They are stored as files with filename <UUID>. You can find the Upload-ID (UUID) in the Debug information of the file after successful upload.

### Large files
Test, if you are able to upload large files. Test if you can pause and resume the upload.

### Add file information
By using the edit button, you can add some file specific information to the upload. Add some information and check if the added info is stored successfully in the <UUID>.info file in the upload folder.
grafik

### Interrupted internet connection
If your connection to the server is interrupted during upload, the upload stopps and can be resumed as soon as you have internt connection again.

### Different servers
If possible try to use different servers (local, hoster, Windows, Linux, ...).

### Different browsers
Try to test the upload with different browsers.


MrMusic:
I have carried out a first test:
Upload to the tmp directory itself works.
The adding to the category and the image processing will probably come in a separate Pull Request?

I have noticed the following things so far:
- In JoomGallery 3.x it was possible to globally set the description, author, published state and access level for all uploaded images. Is that still possible?
- A script file and a CSS file are loaded from https://releases.transloadit.com. Is it possible to save these files locally under media/joomgallery instead of loading them from a foreign server?

Tested with Firefox and Edge with Windows.

Elfangor93:
Global form data submisdion will comelater as well as adding a new record into the database. This PR is only about uploading any file to the server. No further processing of the uploaded data.

Elfangor93:
Global form data submisdion will comelater as well as adding a new record into the database. This PR is only about uploading any file to the server. No further processing of the uploaded data.

Yes the css and mjs file from uppy could also be hosted locally. The drawback: The component would need regular releases since uppy provides new versions every other month. By loading it from servers we dont need to do new releases so regularly.
I guess this is a general discussion to prrform...

reni68:
It works wonderfully.
- Image info edited
- uploaded and renamed
- File content OK
- contain the entered information in the .info

I tested with Firefox and Vivaldi on Linux with XAMMP.
(unfortunately I didn't have any large files "in stock")

PR #70, opened 2023-01-22
Add support for webp images

MrMusic:
This pr adds the webp support as in JG 3.6.0
It is a redo from https://github.com/JoomGalleryfriends/JoomGallery/pull/102 for JG4.
With this it should be possible to upload webp images.

Additionally was fixed the wrong hint 'GD library not found' https://github.com/JoomGalleryfriends/JG4-dev/issues/61

PR #82, opened 2023-03-15
Add warning message instead of fatal error

MrMusic:
Currently, uploading an WEBP file with **'animation'** causes a Fatal Error when using GD as image processor, see https://github.com/JoomGalleryfriends/JG4-dev/pull/70#issuecomment-1450110906
This PR prevents the fatal error and issues a warning message instead.

However, what I don't understand is that if an invalid image type is detected (returning false in the analyse function), normally no record in the database table should be created or? An error in the service?


szepty-ziemi:
Warning  is displayed correctly:

obraz

Nevertheless item is still saved with image linking broken. Shouldn't it rather display error and NOT create new item record at all?
(Or in this ticket only error message is covered?)

obraz

ENV: https://prnt.sc/EIvRdwp8R254

Issue #85, opened 2023-03-30
[FileManager] Thumbnail for "No image" not shown 


Elfangor93:
If you select no image as thumbnail for a category, the default "no image" thumbnail is not shown:
![grafik](https://user-images.githubusercontent.com/39154009/228742424-3377cfc7-e0bd-4ce6-8f78-4bc74c6ad87f.png)

Call Stack:
![Uploading grafik.png…]()

Elfangor93:
Fixed in PR #86

PR #83, opened 2023-03-20
Overhaul of FileManager service

Elfangor93:
## What is this service for

The FileManager service will be used to handle the files and folders created in the filesystem. I offers methods to perform processes (create, delete, check, move) on image files and folders.

**This is a overhaul of PR #26.**

The usage of the FileManager is exactly the same as before but now with a complete new filesystem service. The new filesystem service support the filesystem adapter provider of com_media. Meaning that different filesystems can be used based on which filesystem plugins you have installed.

## What are filesystem plugins

Filesystem plugins are a new type of plugins introduced with Joomla 4. They allow Joomla to communicate with different filesystems. In the Joomla core only one filesystem plugin is preinstalled to do operations with files on the local hosted filesystem:
grafik

There are already several additional filesystem plugins available allowing you to operate with files on other, even not locally hosted filesystems like gDrive, OneCloud, Amazon S3, ...
- https://extensions.joomla.org/extension/amazon-s3-filesystem/
- https://extensions.joomla.org/extension/dpmedia/
- https://www.joomlacontenteditor.net/downloads/plugins/server-filesystem
- https://extensions.joomla.org/extension/google-cloud-storage/

This results that JoomGallery start to support multiple filesystems with this PR without the need of writing code by ourselfs.

## How to use this service

See PR #26

## How to test this service

See PR #26. Everything should work the same as before. You can even use the operations in the GUI to perform the tests.

Affected opertaions:
- create Images
- delete Images
- check Images
- move Images
- copy Images
- create Category
- delete Category
- check Category
- move Category
- copy Category
- get image path
- get category path

Issue #68, opened 2023-01-17
[FileManager service] Problem with images including a dot in the filename


AlexanderSupp:
Please, can somebody test this!
In my tests, Original, Detail and Thumbnail were created successfully. 
But in Tab “Images”, no image will find.

szepty-ziemi:
Hi, Alexander, to which PR your request apply? Or you mean main branch?

AlexanderSupp:
Thanks for the hint. I've now tested both. Numbering yes/no. It's not because of that.

MrMusic:
I suspect the problem with the missing images is because you are trying to upload to a subdirectory. Probably there is a problem with the directory structure, maybe it is related to this problem: https://github.com/JoomGalleryfriends/JG4-dev/issues/44  
As long as this is not solved, you should only work in the top category level.
Furthermore, the created file name depends on whether 'Original filename' is activated in the configuration or not. Although this option is intended for setting the image title in the mass upload, it also plays a role here.

AlexanderSupp:
In my case, it was not the directory.
File Name in the DB Table  was: 20160316162440.jp_20230118_1788119348.jpg
File Name in the Directory was: 20160316162440jp_20230118_1788119348.jpg
The “.” was the problem.
I changed the DB-Table and everything works.

AlexanderSupp:
"Original filename" in my Global Configuration is marked red and set to no.

AlexanderSupp:
![Blumenmer](https://user-images.githubusercontent.com/110494397/215823611-42efbf19-01c3-4be0-b17e-c92841f5fa09.jpg)

Today I was able to reproduce this bug with the photo above.
Details of my config are here.

[DS214Play-20230131-Joomla-Installation.txt](https://github.com/JoomGalleryfriends/JG4-dev/files/10548796/DS214Play-20230131-Joomla-Installation.txt)

I got the same result with WampServer.
Can somebody try this with my picture and config.
Image processing with GD Lib.
Image Title "Blumenmer.jpg-20230131" (look at dot).

![Screenshot 2023-01-31 175307](https://user-images.githubusercontent.com/110494397/215830010-d0ae257f-7622-4792-9fd0-cae3e2fc1911.jpg)
![Screenshot 2023-01-31 175344](https://user-images.githubusercontent.com/110494397/215830023-1a5e7ee9-33e3-418c-ab4c-25be34daf6a4.jpg)
[...]

Issue #56, opened 2022-12-11
[All content types] Titles are not sanitationed


Elfangor93:
When creating new records of any content type (image, category, config set, tag) you need to enter something in the title field. But the value you eneter is not checked neighter sanitationed. This leads to strange titles when you eneter for example

- " " (only a space)
- "<" (smaller than symbol)
- ...

This could be solved by creating a sanitation of the title field value in the Table::bind() method. Similar as we do for aliases.

How does this sanitation should be like?
What do we wanna allow and what should be corrected/cleaned up?

szepty-ziemi:
Joomla does sanitation for title field in articles or categories. Maybe there is a global sanitation function/method that we can use or just copy their solution?

">" is allowed while "<" is not. If title contains "<" it's not allowed to save and title is not cleared.

AlexanderSupp:
So what? Is this a problem for us? Now?

Elfangor93:
Fixed in PR https://github.com/JoomGalleryfriends/JG4-dev/pull/86.
Added the same sanitation as com_content.

PR #86, opened 2023-04-01
Fix upload/filesystem issues

Elfangor93:
This PR fixes some of the issues found in the area of
- upload process
- file management
- sanitisioning

Namely it solves issues #44, #53, #56, #68, #85

This PR was build on top of PR #83. With applying this PR we are much closer to the final single upload process.


eumel1602:
can it be that this PR installation doesn't work?
bildschirmabgriff

AlexanderSupp:
I have exactly the same error-message.

Elfangor93:
The PR is still work in progress...

Elfangor93:
PR is now ready to test...

MrMusic:
Error when saving the configuration:
`Save failed with the following error: Incorrect integer value: 'none' for column ..._joomgallery_configs.jg_msg_upload_type at row 1`
It seems type 'tinyint(1)' is wrong.

MrMusic:
One question about the path field in the categories database table:
The default entry "uncategorised" does not contain a leading slash.
Newly created categories have a leading slash "/".
Which is correct?

reni68:
Tested issues #44 and #85 and both work:
#44: Categories in file system correct and images - old and new - in category Cat-B. 
         All thumbnails visible.

#85: "no image" thumbnail is displayed.

Tested on Linux with XAMPP, Joomla 4.2.9 and PHP 8.0.15.

reni68:
> One question about the path field in the categories database table: The default entry "uncategorised" does not contain a leading slash. Newly created categories have a leading slash "/". Which is correct?

Global configuration can now be saved without error message.

Elfangor93:
>The default entry "uncategorised" does not contain a leading slash.
Newly created categories have a leading slash "/".
Which is correct?

Without leading slash is correct. But a slash does not break functionality. But you pointed me towards some code which could make problems. I am working on it...

eumel1602:
I tested problems #44 #53 #85 #83. Everything works for me now.

What I noticed about the topic: #56 / If I use an image or a category or a tag with only one or more spaces as a title, it saves it, but I can never edit the image again. Saving a e.g. Image with only one space should show an error from my point of view! / when I save a second image with one or more spaces, the system automatically adds a "(2)" and so it's clickable...

bildschirmabgriff

Elfangor93:
@eumel1602 Can you please open an issue for your the titles issue?
I will not fix this here in this PR...

PR #87, opened 2023-04-02
Solve visual problems in image manager

MrMusic:
@eumel1602  has reported here that there is a visual problem in the image manager:
https://github.com/JoomGalleryfriends/JG4-dev/pull/83#issuecomment-1486721378
In some selectboxes the texts are too long and cause a visual problem.
This PR fixes the problem by using the shorter texts as in Joomla itself. See texts in Joomla article manager.

I have tried to change the language constants in the forms/filter_images.xml. Unfortunately, this does not work.

For Category Select Box:
The previous text "Type or Select a Category" is loaded via the language constant JGLOBAL_TYPE_OR_SELECT_CATEGORY. This constant is used in line 71 of .../com_categories/layouts/form/field/categoryedit.php.
To shorten the text I only see the possibility to redefine the constant JGLOBAL_TYPE_OR_SELECT_CATEGORY in the language file com_joomgallery.ini.

For User select box:
For the selection of the owner I used the new constant COM_JOOMGALLERY_FIELDS_SELECT_OWNER in the UserdropdownField.php.

### DIFF
--- a/administrator/com_joomgallery/language/en-GB/com_joomgallery.ini
+++ b/administrator/com_joomgallery/language/en-GB/com_joomgallery.ini
@@ -10,6 +10,7 @@
 ; So, do not translate this string {tip}!
 ; 
 ;Common 
+JGLOBAL_TYPE_OR_SELECT_CATEGORY="- Select Category -"
 JOOMGALLERY="JoomGallery"
 COM_JOOMGALLERY="JoomGallery"
 COM_JOOMGALLERY_CONFIGURATION="JoomGallery: Options"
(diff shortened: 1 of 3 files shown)


administrator/com_joomgallery/language/en-GB/com_joomgallery.ini
```diff
@@ -10,6 +10,7 @@
 ; So, do not translate this string {tip}!
 ; 
 ;Common 
+JGLOBAL_TYPE_OR_SELECT_CATEGORY="- Select Category -"
```
Elfangor93, line 13: I would not overwrite language constants from joomla.ini.
If you want to change this I recommend using the already existing COM_JOOMGALLERY_FIELDS_SELECT_CATEGORY.

Elfangor93: Sorry, you are right its maybe better to overwrite the language constant then overwriting the layout "joomla.form.field.categoryedit".

Elfangor93: By the way I would use the JOPTION_SELECT_CATEGORY constant. This way we dont have to maintain this in the translations.

MrMusic: Thanks for the hint with the layout.
It works well without redefine the constant JGLOBAL_TYPE_OR_SELECT_CATEGORY.

eumel1602:
The text now fits into the box, it works!
What I also noticed is that when I move the mouse over the selection of the category and access level, the mouse pointer is not displayed. see image. The function is OK with the "CLICK".
PXL_20230404_184218893
PXL_20230404_184234258

reni68:
That works for me too.

And I also noticed something else.
The opening windows look different for me.

PR #92, opened 2023-04-17
Dynamic list of configuration options (dropdown-list)

Elfangor93:
This PR adds a general implementation for dynamic configuration options dopdown lists. One application is the selection of available filesystem adapters (plugin type filesystem).

## How is it done
In the XML form definition, add the attribute `dynamic="true"` to the field you want to add dynamic options. The field has to be of a type which supports options like e.g `list` or `radio`. Additionally you have to add the attribute `script="scriptName"` where `scriptName` is the name of the script providing the field with available options.

The script to load the available options can be provided in a plugin listening to the `onJoomGetOptions` event or within the `Joomgallery\Component\Joomgallery\Administrator\Helper\ConfigHelper` where scriptName has to be an available method of this helper class.

## How to test this PR
PR affects only the form view of configuration sets.
Configuration sets form view should work the same as before except for the fact that the field called "Filesystem" now provide a dorpdown list with all available filesystem adapters available in your system.
grafik

You can add new filesystems by install new filesystem plugins or by adding new adapters to already installed filesystem plugins editing the plugin parameters.


reni68:
I get a message when I call up the configuration.

A drop-down field is available.

- see screenshot

Tested on openSUSE Leap 15.4 with Joomla 4.3.0 on XAMPP.

Elfangor93:
Which php version are you using?

reni68:
Am 2023-04-18 23:00, schrieb Elfangor:
> Which php version are you using?
> 

Oh sorry ... PHP 8.0.15.

MrMusic:
My settings:
Joomla 4.3.0
PHP 8.0
Windows 10

>I get a message when I call up the configuration.

I can confirm these warning message in the JoomGallery configuration.

In addition, I tried to test it with the FTP filesystem plugin from Digital Peek: https://joomla.digital-peak.com/documentation/dpmedia/image-services/ftp-plugin
I installed the DP Media Package and the FTP filesystem plugin. These are available for free download.

After configuring and creating the paths, you can select the FTP plugin in the JoomGallery configuration. 👍 
I then created a new category. This category is actually visible in the external FTP filesystem. 👍 
If this category is then opened and closed with "Save & Close", the following error occurs:
error-ftp-plugin

When uploading a image, the same error occurs:
error-upload-image-ftp
However, I am wondering why Filesystem->createFolder() is called here?

Maybe it's also an error in the Digital Peek FTP plugin? because:
- No files and folders from the external FTP file system are displayed in the Joomla Media Manager.
- Uploads of images and creation of folders in the Joomla Media Manager are possible despite error messages, but are not displayed.

Elfangor93:
The FTP plugin from DPMedia is quite strange. I dont understand its functionality not even in the Joomla File Manager.

AlexanderSupp:
I also tried to test the Digital Peek FTP plugin. My goal was to include a directory on my NAS (Synology). However, I failed with my understanding already during the configuration.

Elfangor93:
I think the Digital Peek FTP plugin can only read data from FTP filesystem. As soon as you try to write something it will throw a php exception. Thats probably why it does not work as expected when using it in JoomGallery.

ThomasFinnern:
Hello together,

Surprisingly i did address the exif problem myself the last month.
See:
RSGallery2_J4/ImageExif.php at master · RSGallery2/RSGallery2_J4 · 
GitHub 
<https://github.com/RSGallery2/RSGallery2_J4/blob/master/administrator/components/com_rsgallery2/src/Helper/ImageExif.php>

Points of my concept:

    * items ==> name => value
       for binary data check: if (json_encode ($val) === false){ $val = 
'%binary%'; }
       So i ignore binary values for now :-)

    * Names:
         i keep the type of item as key (EXIF, IFDO, COMPUTED ...) and 
add the name
         $items[$key . '.' . $name] = $val;

    * translations
         I use the name part as part on the translation tag
         $name = explode(".", $ExifTag) [1];
         $translationId = 'COM_RSGALLERY2_EXIF_TAG_' . strtoupper($name);

    * Detect unknown new type/names   from user images
       In RSG2 Maintenance->Exif the user can tell the name of an 
uploaded image
       and the list of items appears. The user can copy this list and 
send it to me.


May be this helps a bit
Regards Thomas





Am 24.04.2023 um 14:30 schrieb Elfangor:

MrMusic:
Probably the php notice can be fixed by changing the ordering of the parameters in line 724 or by changing to `string $needle = '',`
Then this PR can be merged or does anyone have a functioning filesystem adapters plugin?

Elfangor93:
I opened a new issue for the problem with the DPMedia FTP filesystem plugin. It needs a deeper research on which fault it is that the error is thrown. I will research on that later.
Since everything works with the local filesystem plugin we can covering the JG3 functionality at the moment and I therefore think we can merge this PR if there are no other problems found.

MrMusic:
Thank you very much.

PR #97, opened 2023-05-09
Fix PHP Deprecated messages when using php 8.1

MrMusic:
When using PHP 8.1 and Error Reporting = Maximum, there are PHP Deprecated messages.
These warning messages cause e.g. in the image manager that no thumbnails are displayed.

The problem is always the bind function in which ...\strpos($task, 'save')... is used to check for the presence of a string in the '$task' variable.
If $task is NULL, however, this message appears.
Enclosed is a suggestion how to prevent this message. Maybe there is a better suggestion?

### DIFF
--- a/administrator/com_joomgallery/src/Service/IMGtools/GDtools.php
+++ b/administrator/com_joomgallery/src/Service/IMGtools/GDtools.php
@@ -2502,7 +2502,7 @@ class GDtools extends BaseIMGtools implements IMGtoolsInterface
             $bNew = 0;
           }
           $rgbNew = ($rNew << 16) + ($gNew << 8) + $bNew;
-          \imageSetPixel($img, $x, $y, $rgbNew);
+          \imageSetPixel($img, $x, $y, (int) $rgbNew);
         }
       }
     }
--- a/administrator/com_joomgallery/src/Service/IMGtools/GDtools.php
+++ b/administrator/com_joomgallery/src/Service/IMGtools/GDtools.php
@@ -2451,7 +2451,7 @@ class GDtools extends BaseIMGtools implements IMGtoolsInterface
 
           if(($rOrig != $rNew) || ($gOrig != $gNew) || ($bOrig != $bNew))
           {
-            $pixCol = \imageColorAllocate($img, $rNew, $gNew, $bNew);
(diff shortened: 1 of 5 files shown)


Elfangor93:
I will take a free moment to think your changes through. We have to make sure not to change to logic of the function...

administrator/com_joomgallery/src/Service/IMGtools/GDtools.php
```diff
@@ -2451,7 +2451,7 @@ protected function unsharpMask_GD($img, $amount, $radius, $threshold)
 
           if(($rOrig != $rNew) || ($gOrig != $gNew) || ($bOrig != $bNew))
           {
-            $pixCol = \imageColorAllocate($img, $rNew, $gNew, $bNew);
+            $pixCol = \imageColorAllocate($img, (int) $rNew, (int) $gNew, (int) $bNew);
```
Elfangor93, line 2454: I approve this line.

But we have to take care about line 2505 too. I propose doing there the same:
`\imageSetPixel($img, $x, $y, (int) $rgbNew);`

administrator/com_joomgallery/src/Table/CategoryTable.php
```diff
@@ -142,7 +142,7 @@ public function load($keys = null, $reset = true)
 	public function bind($array, $ignore = '')
 	{
 		$date = Factory::getDate();
-		$task = Factory::getApplication()->input->get('task');
+		$task = (String) Factory::getApplication()->input->get('task');
```
Elfangor93, line 145: My proposition, doing the same in a Joomla'ish way:
`$task = Factory::getApplication()->input->get('task', '', 'cmd');`

administrator/com_joomgallery/src/Service/IMGtools/GDtools.php
MrMusic: Thanks for the proposal

administrator/com_joomgallery/src/Table/CategoryTable.php
MrMusic: Thanks for the proposal

Elfangor93:
I tested this PR successfully thank you very much!

PR #102, opened 2023-05-26
Remove updateservers from plugins

MrMusic:
Remove wrong update server because joomla generates a warning when trying to accessing it.


eumel1602:
would like to help Unfortunately I can't understand what to do here. Is there an incorrect server entry in the update sources? When exactly and where should I select the update server? Why change? Is there an error description?

AlexanderSupp:
I can't produce a warning. However, there is no harm in changing something here.
Screenshot 2023-05-29 131246

Screenshot 2023-05-29 131435

MrMusic:
>Is there an incorrect server entry in the update sources?

Yes

>I can't produce a warning

Testing instruction:
Change the error reporting in the Joomla configuration to 'Maximum'.
In the Joomla backend, go to Extensions Update. Click on 'Check for updates'.
Check you php error.log for entries that contain the url `https://www.component-creator.com/index.php?task=builder.preupdatecheckhook&option=com_combuilder&component=NzMzOTAtMjAxMTE0`
These update-url is currently entered during the installation by the additional plugins (finder, webserices, privacy).
It is not an important thing, but since it is no longer correct, it should be removed.

AlexanderSupp:
Thank you for the hints. I tested it and found a lot of information in the log files. It was quite confusing for me. I have therefore canceled further activities on this. Thanks again for the explanations.

eumel1602:
Where is this log file? In the root of my website, the error_log is empty.

AlexanderSupp:
These log files are outside the joomla files. See example below. Take a look at php_error.log.
Screenshot 2023-05-30 213814

MrMusic:
>Where is this log file? In the root of my website, the error_log is empty.

Not the Joomla log. It's about the error log from your web server.

What kind of installation do you use? Hoster? XAMPP? Wampserver?
Normally there should be a button in the configuration interface to view the log file.
With XAMPP e.g. here:
xampp-error-log

eumel1602:
Is this PR still up to date?
I use hoster (alfahosting) for testing. I can't find anything to display the error logs except for the error logs (Log folder) on the server that I find via FTP?! '

Maybe there is more help here...

MrMusic:
The log entries are not in the Joomla log. I don't know if and how you have access to the server logs at alfahosting. Here it would be best to test locally.
This PR does not change anything in the JoomGallery. It only prevents the wrong update server for the plugins from being entered in Joomla when installing the JoomGallery. You can click on updates sites in Joomla and see which urls are entered there. If you check the urls, you will notice that the ones for the JoomGallery plugins are incorrect.
But actually it is not important at all.

Elfangor93:
I recommend to not remove the updateservers right now. The errors in the log does not disturb us from developing but we see in the code that we have to consider the updateservers. Its a reminder for us.
Therefore I propose to close this PR.

MrMusic:
>...but we see in the code that we have to consider the updateservers. Its a reminder for us

If the corresponding plugins keep being installed and uninstalled together with the component, no updateservers are necessary at all. Entering an updateserver in the future only makes sense if the extension is installed, uninstalled and updated independently.

Elfangor93:
We will serve some very minimal plugins with the component. This plugins will not have their own updateserver.

MrMusic:
Update this PR to the current plugins.

Elfangor93:
Changes were included in PR #287

PR #104, opened 2023-06-05
Access service

Elfangor93:
## What is this service for

The Access service will be used to extend the default ACL logic of Joomla core in order to allow checks against inown and own access rules.
Inown and own access rules are used in JoomGallery to manage ACL exceptions. For example, a user may perform a certain task within an asset of which he is the creator (owner), even if this task is generally forbidden for him.

## How to use this service

You can load the service anywhere in Joomla by doing

```
$component->createAccess();   // instantiate the service
$acl = $component->getAccess();   // load the service into your scope
$acl->setUser(<userID>); // perform checks against a specified user. If no user is set the current user will be used.
$res = $acl->checkACL(<action>, <asset>, <ID>); // check if the user is permitted to perform <action> on <asset> with the ID <ID>
```

The following actions are available:
`add, admin, delete, edit, editstate, manage`

The following assets are available:
`image, category, tag, config`

## How to test the service

Open the file `/administrator/com_joomgallery/src/View/Faulties/HtmlView.php` and insert the following code on line 62 after `throw new \Exception(implode("\n", $errors)); }`

Different test instructions see comments below...


Elfangor93:
### Test Methodology

\<USERID>: User ID (integer) - Choose a user for which to perform the ACL checks
\<ACTION>: The action you want to perform the ACL check (string)
\<ASSET>: The asset for which you want to perform the ACL check (optional, string, default: "")
\<ID>: The id of the asset (optional, integer, default: 0)

```
$this->component->createAccess();
$acl = $this->component->getAccess();
$acl->setUser(<USERID>); // optional
$res = $acl->checkACL(<ACTION>,<ASSET>, <ID>);
dump($res);
```

### Test Instructions

1. Apply the the code from above as described in the first comment.

2. Prepare component and settings
- Create a user (additional to the Super user)
- Create a category (additional to default category)
- Create at least one asset each (image, category, tag, config-set)
- Apply permission options accordingly
- - Global component permission options
- - Category permission options
- - Asset permission options

3. Perform the test by reloading the maintenance view and check if the output is correct.
false: permission denied
true: permission accepted

szepty-ziemi:
Not directly related with feature test, but while uploading file to create test date I get an error:
obraz

Installed zip from acl-apply branch.
@Elfangor93 please advise.

PR #105, opened 2023-06-05
Fix approve button

Elfangor93:
Before this PR the approve button did not work korrectly. It was not possible to unapprove an image once it was approved using the images view.

With this PR this should work now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@
 /.idea/fileTemplates/includes/PHP Property Doc Comment.php
 /.idea/inspectionProfiles/Project_Default.xml
 /.idea/vcs.xml
+node_modules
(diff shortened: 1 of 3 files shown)


eumel1602:
Maybe I'm misunderstanding it but:
I can click the "Approved" button in the general image view (Image 1). This works (has always worked), if I edit in the single image I can't do that, there is no selection area (image 2)
If the "approved" button is to be regulated differently, please provide a brief description here
(Win 10 / J4.3.2. PHP 8.0
veroeffentlicht1
veroeffentlicht2

Elfangor93:
Unapproving images in the lost view didnt work before this PR.

eumel1602:
ok, i couldn't find any mistake

MrMusic:

Issue #90, opened 2023-04-05
Better metadata support


Elfangor93:
It would be huge if JoomGallery could add better support for image metadata. Currently JoomGallery uses the built in function [exif_read_data()](https://www.php.net/manual/en/function.exif-read-data.php) and iptcparse(https://www.php.net/manual/en/function.iptcparse.php) from the GD library to read metadata.
In most of the cases this is good enough.

But with an external tool we could improve the functionality and compatibility of metadata handling like we do with the external tool ImageMagick for image manipulation.
Advantage: More functionality.
Disadvantage: Tool has to be installed on the server.

One of the two tools I could think of using:
- [ExifTool by Phil Harvey](https://exiftool.sourceforge.net/) written in PEARL
- [The PHP JPEG Metadata Toolkit](https://www.ozhiker.com/electronics/pjmt/) written in PHP

Does anyone would have time to test these two tools?

AlexanderSupp:
I use https://exiftool.org/ by Phil Harvey to control all things around EXIF & Co.
It's easy to install on Windows and Mac.  I also use it with WampServer.
It's part of my provider environment, as a standard when ImageMagick also is available.
It's installed on my both Synology-Machines.
ExifTool by Phil Harvey is the most used tool for Metadata Management.
In addition, there is a graphical user interface ExifToolGUI for easy use.
https://www.computerbild.de/download/ExifToolGUI-3329098.html
https://www.heise.de/download/product/exiftoolgui-63032
 
![Screenshot 2023-04-05 113805](https://user-images.githubusercontent.com/110494397/230044197-3c2a1e5a-7645-4558-bb88-4f420cf328b8.png)

To control Metadata, a tool called Exif Pilot is also useful.
https://www.colorpilot.com/exif.html

I have no experience with The PHP JPEG Metadata Toolkit. But I remember an image viewer with this layout.
I try to find out more.

AlexanderSupp:
I took a look at https://www.ozhiker.com/electronics/pjmt/
Doesn't inspire confidence in me.
Last changes at least 8 years ago.
Haven't found any information about the use.

I prefer the usage of ExifTool.

AlexanderSupp:
Now I'm sure that ExifTool is the best choice.

See https://www.linux-magazin.de/news/digikam-7-3-holt-exiftool-an-bord/
_"DigiKam 7.3 holt ExifTool an Bord
[...]

PR #109, opened 2023-06-17
Recreate images

Elfangor93:
This PR adds the functionality to recreate imagetypes (thumbnail, detail, original, ...) based on the current parameters in the config sets.
In the Images view there is now in the "images processing" dropdown a button "Recreate".
grafik

# How to test
- Change the settings in the configuration set
- Select all images for which you want to recreate the imagetypes
- Press the Recreate button


AlexanderSupp:
After recreate of a single image, without changing anything, the original image has the dimensions of the detail image.

MrMusic:
>After recreate of a single image, without changing anything, the original image has the dimensions of the detail image.

Confirmed. After recreate the dimension of the original image is wrong.

Elfangor93:
>After recreate of a single image, without changing anything, the original image has the dimensions of the detail image.

I can not reproduce this issue.

Can you please take a look into your database and see if in the table #__joomgaller_img_types, the record with typename `original` has a param called `jg_imgtype` and it has the value `1` ?

The only reason the script takes the detail image as source for recreation is if this param is missing or has a value different than 1. This would result that the new original image is a copy of the old detail image and would have the same dimension as the detail.
This happened to me during testing when this param for mysterious reasons disappeared. Then I had to save the configuration set once more and it worked again as expected.

Elfangor93:
The commit #9a6c16e will force to use the original image as the source image even if the corresponding parameter in the database is missing or corrupt.
This fix does not really solve the problem since the real problem is the missing or wrong parameter value in the database...

AlexanderSupp:
I uninstalled JG4 and I installed recreate-imagetypes only.
Then I tested with only 2 images.
The parm "jg_imgtype":"1" is in detail and in thumbnail, but not in original. See screenshot below.
However, now I can't reproduce this error as before.
Screenshot 2023-06-18 192818

I will continue my test now.
Am I correct in saying that restyling an image doesn't affect the metadata?
Or will this data also be newly provided?

AlexanderSupp:
OK, I tested it. The recording date is not changed. So I assume metadata is fundamentally not restored.

AlexanderSupp:
I tested the resizing of detail, thumbnails and (Warning) the originals. Also, the watermark. Everything works well.
Works as designed. If you change the original and if you have no backup, your photo will never be restored. At this point, a warning make sense.
It would be fine if in a later phase the metadata were also recreated. See Issue #90
Anomalies:
In the Images screen, the thumbnails displayed with a frame. This doesn't look nice, except you have a square. Just omit the frame.
The watermark in a thumbnail is cut.

From my point of view, this PR is ready to merge into main.

MrMusic:
>...This happened to me during testing when this param for mysterious reasons disappeared....

Yes, the parameter is missing in the database record.
If the configuration is saved without 'Activate Imagetype' being enabled (sensitive setting), 'jg_imgtype' is lost.
If 'Activate Imagetype' was enabled then the value is saved.
This only affects the 'original' image type. Maybe it has to do with the fact that it is a sensitive setting? 
The problem does not occur with newly created image types.

Elfangor93:
>This only affects the 'original' image type. Maybe it has to do with the fact that it is a sensitive setting?
Yes. Sensitive input fields gets the form field attribute `disabled` which prevent its values beeing submitted: https://stackoverflow.com/questions/1355728/values-of-disabled-inputs-will-not-be-submitted

With the lastest commit I added code such that jg_imgtype is always retained and even reset if it was once lost.

MrMusic:
>With the lastest commit, I added code such that jg_imgtype is always retained and even reset if it was once lost.

The fix always sets the value to 1. Therefore it is no longer possible to deactivate the 'original' imagetype (Value 0).

MrMusic:

PR #110, opened 2023-06-17
Replace imagetype

Elfangor93:
This PR adds the functionality to replace/exchange a specific imagetype (thumbnail, detail, original, ...). This is useful for example, if you are not satisfied with an automatic processing result and you want to create the image for this imagetype locally by hand and then replace it.
In the Image edit view there is now a "Replace" button in the preview modal of each imagetype. This button brings you to a from view where you can select/upload the image used for replacement.
grafik

Replacement form view
grafik
The left side is to be filled out. The right side gives you the info on which image you are working on.

### DIFF
--- a/administrator/com_joomgallery/forms/image.xml
+++ b/administrator/com_joomgallery/forms/image.xml
@@ -260,6 +260,22 @@
            type="hidden"
            filter="stirng" />
 
+    <field name="replacetype"
+           type="jgimagetype"
+           label="COM_JOOMGALLERY_IMAGETYPE"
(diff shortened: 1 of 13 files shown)


AlexanderSupp:
I miss a close button here.
Screenshot 2023-06-19 013722

MrMusic:
It looks good to me. 
Only the following PHP warning appears during the replacement:
`PHP Warning:  Undefined array key "filename" in ...administrator/components/com_joomgallery/src/Service/Uploader/SingleUploader.php on line 117`

MrMusic:

PR #119, opened 2023-07-24
Fix mass upload with uppy

MrMusic:
Fix for issue https://github.com/JoomGalleryfriends/JG4-dev/issues/115
Then the mass upload with uppy works should work again.

PR #121, opened 2023-07-28
Use 'Admin approval' setting for uploads

MrMusic:
Actual result:
So far all uploaded images are automatically not approved.

Changes:
Move and activate the option 'Admin approval' to the configuration tab 'General Settings -> Uploads'. 
Modification of the language string.
Consider the setting for the uploads.

Expected result:
Depending on the setting in the configuration, the images are automatically approved or not when uploaded.


MrMusic:
Thank you for testing.

>PS: Multiple Upload is not working here.

That is correct, as this is contained in a separate[ pull request](https://github.com/JoomGalleryfriends/JG4-dev/pull/119). Unfortunately, this has not been merged yet. 
Since the change only affects a single line, you can make the change yourself if necessary.

administrator/com_joomgallery/src/Service/Uploader/Uploader.php
```diff
@@ -204,6 +204,12 @@ public function retrieveImage(&$data, $filename=True): bool
       }
     }
 
+    // Image approved by default
```
Elfangor93, line 207: I would rather add this lines to the ImageTable::bind() method.
(Joomgallery\Component\Joomgallery\Administrator\Table\ImageTable).
When in the ImageTable the configuration `jg_approve` will be considered everytime an image gets created. When in the uploader service it is considered only when adding an image using the uploader service.

administrator/com_joomgallery/forms/config.xml
```diff
@@ -157,6 +157,17 @@
         <option value="2">COM_JOOMGALLERY_CONFIG_REPLACE_SHOWWARNING_HINT</option>
       </field>
 
+        <field name="jg_approve"
```
Elfangor93, line 160: If you add a new configuration setting you also have to add the corresponding column into the database table (#__joomgallery_configs). Otherwise the setting will not be saved.

Elfangor93 (changes requested):

administrator/com_joomgallery/src/Service/Uploader/Uploader.php
MrMusic: Should be fixed now.

rowi68:
Hello everyone,
just finished my first test... :-)

So it works for me on a subdomain...,
cycling

....but I can't save images local. I don't know, perhaps a configuration issue with Xampp? (I use Xampp first time)
unsupported_image-filetyp

AlexanderSupp:
@rowi68 
....but I can't save images local. I don't know, perhaps a configuration issue with Xampp? (I use Xampp first time)

Please check in Joomla
Content, Media, Options, Allowed Extensions and Legal Image Extensions,
and also
JoomGallery, Configuration Sets, Supported image types
Maybe you use a jp**E**g-file and only jpg is defined.

administrator/com_joomgallery/src/Table/ImageTable.php
```diff
@@ -157,6 +157,7 @@ public function bind($array, $ignore = '')
 	{
 		$date = Factory::getDate();
 		$task = Factory::getApplication()->input->get('task', '', 'cmd');
+		$app  = Factory::getApplication()->bootComponent('com_joomgallery');
```
Elfangor93, line 160: Please dont use the variable name `$app` for the JoomGallery extension object. Everywhere in Joomla `$app` is used to hold the CMSApplication object.
As a proposition, I am using `$component` in the code as the storage variable for the JoomGallery extension object.

Elfangor93 (changes requested):

rowi68:
Hello Alex,

Thank you for advices. I could not find the reason. All file types like jpg,jpeg,png,gif,webp are defined and allowed, but non of them are storable. Because it works on my subdomain, I will continue testing there in future....

MrMusic:
>....but I can't save images local. I don't know, perhaps a configuration issue with Xampp? (I use Xampp first time)

@rowi68   thanks for testing.
If you are using Xampp for the first time, the GD extension is probably not activated. You have to activate it in php.ini.
Check which GD version is displayed in the JoomGallery configuration.

administrator/com_joomgallery/src/Table/ImageTable.php
MrMusic: @Elfangor93 thanks for the assistance and explanation.
I hope it fits now.

Elfangor93 (approved):

rowi68:
@MrMusic This was the cause. Now it works. Thanks for help....

Elfangor93:
I tested this PR successfully. Tank you very much!

Issue #117, opened 2023-07-13
Error displayed when trying to check-in configuration set


szepty-ziemi:
Steps to replicate:

1. Go to configuration sets
2. Edit a configuration set
3. Click "Go back" in the browser to return to configuration sets list
4. Now configuration set will have a padlock
5. Click on a padlock to check-in configuration set

Result:

User is redirected to images list with error message "Check-in failed with the following error: "

![obraz](https://github.com/JoomGalleryfriends/JG4-dev/assets/60470150/64cccb71-6042-4464-85e1-08b08ed3c46c)

Expected result:

Configuration set is checked-in, user stays on configuration set lists, message confirms that check-in was successful.

![obraz](https://github.com/JoomGalleryfriends/JG4-dev/assets/60470150/91afaa04-9c0d-47e5-b13d-4f03595d570c)

ENV: https://prnt.sc/jBazGjWsKpT1

reni68:
Yes, here with me also so and it happens a change in the menu from "Configuration" to "Images".

MrMusic:
Closea as completed

PR #122, opened 2023-07-28
Fix for configuration check-in

MrMusic:
Fix for issue https://github.com/JoomGalleryfriends/JG4-dev/issues/117


reni68:
It works.
The configuration set is checked in, you remain in the list of configuration sets.
grafik

eumel1602:
I can confirm that the bug has been fixed. This works for me!
Win 10/ Test Environment Online / J 4.3.3. / PHP 8.0.22

PR #125, opened 2023-08-11
Add Control Panel

MrMusic:
This PR adds a 'Control Panel' to the admin area of the JoomGallery.
In this control panel important information for the user should be displayed.
This is only a draft for now. 

What information should be displayed?
How can the user experience be improved?
Any other suggestions?

Please note:
During installation an administrator module is created to display a feed from the JoomGallery website.
Unfortunately there is a bug in Joomla that currently leads to an endless loop.
A [bugfix](https://github.com/joomla-framework/filter/pull/50) was already submitted a long time ago but unfortunately not merged. 
In order to test the control panel with the feed module, you have to make the following change in the file /libraries/vendor/joomla/filter/src/InputFilter.php at line 728:
`$nextAfter = $nextBefore + $closeQuoteChars;`
Presumably it would make sense to submit a PR for the framework again to draw attention to the problem?


eumel1602:
Thank you for your effort.
I just installed it, didn't change any files and so far I haven't gotten an error message in BE. (Where should it come from??) The few clicks all worked
My thoughts on the CP:
- The possibly used language (version) should be displayed in the notification area
- In the notification area, the best ratings, the pictures with the most comments or the last comments should be displayed in a module
- Under Statistics there should be an explanation about the "X" and "Hack".
- From each sub-item (images, cat, settings, ...) there should be a back button to the control panel
- The news area as in JG 3.6. I like it! should go back in
- I consider the following buttons or direct links to be important and should be in the upper area: Cat. / Pictures /Comment / Rating / Tag / Confirmation / and MORE (where other things are then displayed with a submenu, e.g. maintenance manager or mass moving images ect.)
I don't know if that's possible, maybe it's possible to display the folder structure or call it up via the panel...

rowi68:
The Feed Module (if  you mean "JoomGallery News" )  is working well without error message. Only Link to [Joom::Gallery friends - Startseite](https://www.joomgalleryfriends.net/?format=feed&amp;type=rss) is not correct.

Possible Improvements CP:
- I also missing a back button to the control panel, this would be very helpful
- display in statistic (or somewhere else) e.g.  top downloads, popular images, recent upload, best rated
- yes, a folder structure would be nice...
- display all available extensions, maybe separated in  activ/inactiv/not installed

AlexanderSupp:
This Link
 Joom::Gallery friends - Home
 downloads a file called _qVwXkfZ. Is this the final state.
 I assume that the links to the German version will come later.
 From my point of view, here is not the place for a folder structure.
 A folder structure will need everywhere, and it would be great to implement that as a module.
All other things are mentioned above.
This panel is a good idea and has room for improvement.
Thank you.

rowi68:
I would expect when a link text  describes "to home page" to come there and not open a feed file. Otherwise the text is wrong...

MrMusic:
Thanks you for your advice.

>...installed/available extensions...

I have tried to display the installed extensions. The extensions that have 'joomgallery' in their name or are installed in the 'joomgallery' folder are displayed.
However, it is possible that not all extensions are found.


>Add Button to control panel

The desired button back to control panel has now been added.


>...display the folder structure...

I can't imagine what this should look like and where it should be displayed.
Maybe someone could create a picture of how and where this should be displayed in CP.
What would be the use case for this in the control panel? Wouldn't such a display in the category or image manager make more sense?


>The Feed link to Joom::Gallery friends - Home

The link is generated automatically by Joomla. This was also the case in Joomla 3.x. Unfortunately, I see no possibility to change this, only to switch the title off.
You can try it out by creating a feed module yourself in the admin area. For example, you can enter https://www.joomla.org/announcements.feed?type=rss as the feed address.


>I haven't gotten an error message in BE. (Where should it come from??)
The Feed Module (if you mean "JoomGallery News" ) is working well without error message.

Confirmed, but it is strange.
When i started with the Controlpanel recently, the problem was still there. It seems to me that the problematic multibyte character may no longer occur?


>...display in statistic (or somewhere else) e.g. top downloads, popular images, recent upload, best rated...

I am still looking for a solution...

rowi68:
The back button is working. Thank you.

Feed link it is really not a problem... I had not realised that the feed is automatically created by joomla.

Folder structure I thought maybe similar like:
structure_2
or
structur_1

But you are right, in CP is not right place for that. Should be in categorie manager (or like mentioned as module)...

MrMusic:
Images top lists are now added. Note: Before anything can be displayed, you must enter a few values in the database table.
Added important PHP settings.

AlexanderSupp:
Why do we now have the "<- Control Panel" button everywhere in "Images" to "Maintenance"?
To my understanding, this is not the rule.
Discussion point for next meeting?
If you like to go back, every browser has this function.
What does "<-" mean if I come from "Tags" for example?
On the other hand, I can jump anywhere at any time using the left navigation panel.

eumel1602:
> Why do we now have the "<- Control Panel" button everywhere in "Images" to "Maintenance"? To my understanding, this is not the rule. Discussion point for next meeting? If you like to go back, every browser has this function. What does "<-" mean if I come from "Tags" for example? On the other hand, I can jump anywhere at any time using the left navigation panel.

This is exactly what rowi68 and I asked for. I think this thing is mandatory. You should not rely on the browser BACK function here. people will look up at the options how do I get back now....
Let's ask the other way around. what disadvantage should this area have!
Mr Music, thanks for adding the new features...

eumel1602:
> I like the control panel. I noticed that this edit box (see screenshot) is not displayed correctly. It "hides" behind the menu. grafik

Where exactly is that supposed to be?

eumel1602:
Should this logo be installed in the header area of the control panel?
bildschirmabgriff

rowi68:
I have the same issue with edit button of feed module and also the hide button is not working. Maybe you could place edit box on right side?

2023-08-20_07h52_43

rowi68:
In my opinion would be "statistic" and "information" only a side information and should placed at bottom of page.

MrMusic:
Add css class to move cogs icon to right.
Add Donation area at bottom.

MrMusic:
>In my opinion would be "statistic" and "information" only a side information and should placed at bottom of page.
Should this logo be installed in the header area of the control panel?

What exactly should be at the top?
Maybe we should create a new logo or something similar?

rowi68:
Cog icon on the right side is OK now and edit box visible. But "hide/unpublish" button is not working. I am not sure if modul should permanently visible or not. If permanent, then it would probably better to remove the hide/unpublish button..
2023-08-21_20h05_58

The donation area is in my opinion not visible enough. Missing some color/highlighting for better attention. What do you think if we would integrate the logo in donation area and install in top/header position.
I could imagine follow order:
   1. Logo + Donation
   2. Ranked images (thank you for accordion menu --> I like it :heart_eyes:)
   3. Feed module (permanent visible?)
   4. Statistic & Info
   5. Installed extensions and others....

eumel1602:
> Maybe we should create a new logo or something similar?

No, I think this logo is known for now and shouldn't be changed at the moment, unless it gets a fresh cure. The users shouldn't be confused by what's going on here

reni68:
My Cog icon it's ok.
Is it possible to set the number of displayed images for these areas (see screenshot) somewhere or add such a setting?
grafik

MrMusic:
>Is it possible to set the number of displayed images for these areas (see screenshot) somewhere or add such a setting?

The maximum number of images is currently defined in line 60 of the file administrator/components/com_joomgallery/src/View/Control/HtmlView.php:
`$imglimit      = 10;`
What number makes sense? is it necessary to make the number changeable in the configuration?

reni68:
> is it necessary to make the number changeable in the configuration?

Yes, I would like that.

MrMusic:
In order to 'unpublish' the administrator module, additional events would be necessary, so I have only hidden the 'unpublish' via css.
I changed the order of the toplosts within the control panel and added the logo and an additional info text at the top.
The number of images shown can now be set in the configuration.

eumel1602:
Thanks for the changes.
When I go to the config, I can see under general settings: "Display Images in Control Panel", but the value remains the same after saving?! Doesn't that have any effects yet?
The Control Panel is currently completely untranslated. Do the strings only come when the PR has worked in the main? How was it?

MrMusic:
>When I go to the config, I can see under general settings: "Display Images in Control Panel", but the value remains the same after saving?!

Have you made a new installation of the branch?
Because the database field is only created during a new installation.


>Do the strings only come when the PR has worked in the main? How was it?

Yes, the new language strings only appear for translation after the PR has been merged into the main branch.
It is not sure yet whether this will ever happen.

eumel1602:
> > Have you made a new installation of the branch? 

I've set everything up again and now the setting works! Note: A setting of "0" displays "all" new images.

rowi68:
> The number of images shown can now be set in the configuration.

Sorry, can't find it. Could you tell me the path to come there?

I think, for optical separation it would be nice to have a larger distance or horizontal rule  (or both) between toplists and logo/donation..

eumel1602:
> Sorry, can't find it. Could you tell me the path to come there?

Globale conficuration - generell setting - (6.point) -> "Display Images in Control Panel",

reni68:
Thanks for the changes.

MrMusic:
>I think, for optical separation it would be nice to have a larger distance or horizontal rule (or both) between toplists and logo/donation..

A horizontal line added.

>Note: A setting of "0" displays "all" new images.

I have added an additional check so that only between 1 and 100 images can displayed.
Larger values would generate a high server load, as all top lists are recreated every time the control panel is called up.

reni68:
I see this message in the ControlPanel:
grafik

Joomla 4.3.4 auf Linux und Xampp mit PHP-Version 8.1.17

rowi68:
Thank you for changes. Looks better with horizontal line now...

I think that this text :
>JoomGallery is one of the most comprehensive and largest image gallery components for the CMS Joomla!
Published under the GNU/GPL license.JoomGallery is developed and maintained by the JoomGallery friends.
 
under the logo would be not very important for backend-users, becauce JG4 is already installed and webmaster knows that usually. Would it not be better to place the logo above donation text and display  one box only?
What do you think?

MrMusic:
@reni68 @rowi68 :
You are both right. It depends on whether the 'Feed Date' is activated in the administrator module settings or not.
Unfortunately, I don't know what I can do about this warning message. I would recommend to deactivate the Feed Date. By the way, a similar message appears when you create a new image in backend.

reni68:
> I would recommend to deactivate the Feed Date. 

Yes, with the feed date deactivated, this message does not appear.

rowi68:
I cannot reproduce the error. No matter if i activate or deactivate the feed date --> everything is fine...

rowi68:
I cannot reproduce the error. No matter if i activate or deactivate the feed date --> everything is fine.

reni68:
I have installed this branch in another Joomla installation (basic data are identical) and can no longer reproduce the error. Not bad either ;-)

MrMusic:
>Would it not be better to place the logo above donation text and display one box only?

A single box would go across the entire width of the screen. Would it look good if the logo and the text went across the entire width of the screen?
Here is a screenshot how it looks like:
control-panel

reni68:
too big for my liking :/

eumel1602:
Yes, I also think it's too big

MrMusic:
Small changes:
Fix merge conflicts
Remove administrator feed module when uninstalling the component
Remove button Maintenance

rowi68:
I had not intended the logo to go over the whole width :-)
In my opinion it looks better now with only one box, but of course logo smaller (maybe 50-60% width only)

MrMusic:
@rowi68 Thank you for your suggestion.
However, I don't want to make any more changes before the Alpha 1 release. I assume there will be more changes after the release.

AlexanderSupp:
I think its time now to merge this PR into main.

MrMusic:
>I think its time now to merge this PR into main.

@Elfangor93  What do you mean?

script.php
```diff
@@ -686,51 +726,53 @@ private function installPlugins($parent)
...
-							'type LIKE ' . $db->quote('plugin'),
-							'element LIKE ' . $db->quote($pluginName),
-							'folder LIKE ' . $db->quote($pluginGroup)
-						)
-					);
-				$db->setQuery($query);
-				$db->execute();
-			}
-		}
+    if(!$plugins || empty($plugins->children()) || count($plugins->children()) <= 0)
```
Elfangor93, line 729: I guess you are reverting some code that was changed/optimized  in one of the last PR's.

script.php
```diff
@@ -780,37 +822,35 @@ private function installModules($parent)
 			$modules = $parent->get('manifest')->modules;
 		}
 
-		if (!empty($modules))
-		{
+    if(!$modules || empty($modules->children()) || count($modules->children()) <= 0)
```
Elfangor93, line 825: Same here...

script.php
```diff
@@ -833,45 +873,47 @@ private function uninstallPlugins($parent)
...
-						$app->enqueueMessage(Text::sprintf('COM_JOOMGALLERY_SUCCESS_UNINSTALL_EXT', 'Plugin', $pluginName));
-					}
-					else
-					{
-						$app->enqueueMessage(Text::sprintf('COM_JOOMGALLERY_ERROR_UNINSTALL_EXT', 'Plugin', $pluginName), 'error');
-					}
-				}
-			}
-		}
+    if(!$plugins || empty($plugins->children()) || count($plugins->children()) <= 0)
```
Elfangor93, line 876: Same here...

script.php
```diff
@@ -901,62 +943,55 @@ private function uninstallModules($parent)
       {
         $modules = $parent->get('manifest')->modules;
       }
+    }
 
-      if(count($modules->children()))
+    if(!$modules || empty($modules->children()) || count($modules->children()) <= 0)
```
Elfangor93, line 948: Same here...

Elfangor93 (changes requested):

Elfangor93:
Everything looks good except fot the installation script file. You are reverting here some stuff I optimized lately.

script.php
MrMusic: i have updated my branch. hope it is correct now.

script.php
MrMusic: i have updated my branch. hope it is correct now.

script.php
MrMusic: i have updated my branch. hope it is correct now.

MrMusic (commented):
i have updated my branch. hope it is correct now.

Elfangor93 (approved):

Issue #127, opened 2023-08-14
Error on install -> deinstall


ThomasFinnern:
Had an error on previous installation upload images (version git 13.08.2023) 
So i tried to uninstall and begin from scratch. Had an error uninstall. So i did setup 
J4x and database anew and did just install and then uninstall direct afterwards

i got same error again. 

![2023-08-14 20_56_26-Error_ 0](https://github.com/JoomGalleryfriends/JG4-dev/assets/7040580/4ad7ebc6-b3b1-4e76-b82b-f1bc21a8c90b)

MrMusic:
Confirmed.
Thanks for reporting this.

PR #128, opened 2023-08-15
Fix error during uninstallation

MrMusic:
Fix for issue https://github.com/JoomGalleryfriends/JG4-dev/issues/127

Thanks for reporting this.

### DIFF
--- a/script.php
+++ b/script.php
@@ -902,7 +902,7 @@ class com_joomgalleryInstallerScript extends InstallerScript
         $modules = $parent->get('manifest')->modules;
       }
 
-      if(count($modules->children()))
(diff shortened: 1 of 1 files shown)


ThomasFinnern:
Tested successfuly

eumel1602:
test ok

MrMusic:
@Elfangor93  can this PR merged?
I think it would be important to be able to uninstall the component for further testing.

Elfangor93:
Works principially. But it intruduces redundant code.
With this work around `empty($modules->children())` gets checked double. PR #129 solves the same issue but does it with a bit nicer code.

PR #129, opened 2023-08-21
Fix error during uninstallation v2

Elfangor93:
Fix for issue https://github.com/JoomGalleryfriends/JG4-dev/issues/127

Solves the same problem like PR #128 but removes redundant code parts.
This PR also solves the same issue also in other methods of the installer/uninstaller script.


rowi68:
Tested ✅ successfully.

eumel1602:
Worked. I was able to uninstall it with no problems

WIN 10 / PHP 8.0.x / J4.3.4 /

reni68:
it works.

MrMusic:
The uninstallation of the component now works without any problems.
However, I noticed that when I reinstalled the component the plugins finder, webservices etc. were no longer installed. Is this a bug or is it intentional?

MrMusic:
Works now without problems. Thank you.

PR #132, opened 2023-09-03
Code adjustments for Alpha 1 release

Elfangor93:
This PR prepars the code fot the first alpha release.
It adds a message that the component is still in development phase and intended for testing only.

Release date is set to be the 10.09.2023

PR #133, opened 2023-09-05
prevent php deprecated warning

MrMusic:
Fix PHP deprected warnings:
```
PHP Deprecated:  explode(): Passing null to parameter #2 ($string) of type string is deprecated in \administrator\components\com_joomgallery\src\Service\Filesystem\Filesystem.php on line 831
PHP Deprecated:  explode(): Passing null to parameter #2 ($string) of type string is deprecated in \administrator\components\com_joomgallery\src\Service\Filesystem\Filesystem.php on line 839
PHP Deprecated:  explode(): Passing null to parameter #2 ($string) of type string is deprecated in \administrator\components\com_joomgallery\src\Service\Filesystem\Filesystem.php on line 847
```

It tries to read values from the configuration that are not existent.
The code is only commented out so that it can be easily reactivated later.
See also https://github.com/JoomGalleryfriends/JG4-dev/pull/132#issuecomment-1705452052

### DIFF
--- a/administrator/com_joomgallery/src/Service/Filesystem/Filesystem.php
+++ b/administrator/com_joomgallery/src/Service/Filesystem/Filesystem.php
@@ -820,7 +820,7 @@ class Filesystem implements AdapterInterface, FilesystemInterface
       $images = array_map(
         'trim',
         array_filter(
-          explode(',',$this->component->getConfig()->get('jg_imagetypes',null)),
+          explode(',',$this->component->getConfig()->get('jg_imagetypes', '')),
           fn($value) => !is_null($value) && $value !== ''
         )
       );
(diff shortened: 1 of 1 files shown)


reni68:
These messages no longer appear. :+1:

Elfangor93:
I don't like this approach of commenting out stuff to get rid of php warning.

It's just a warning and will not disturb any users since they will not have php error reporting to the maximum. But it helps the developer not to forget that there is an issue to be fixed. Therefore I propose o close this PR.

MrMusic:
The testers are explicitly advised to set the error reporting to 'Maximum'.
Therefore i would delete the unnecessary code.

Erftralle:
I wonder if JoomGallery will also include the management of audio files, documents and videos in the stable release? To get rid of the warning you could just use an empty string as the default parameter instead of NULL.

MrMusic:
>I wonder if JoomGallery will also include the management of audio files, documents and videos in the stable release?

At the moment it is not planned. 
But i would like it if JoomGallery also supported videos (and others) at some time.

>To get rid of the warning you could just use an empty string as the default parameter instead of NULL

Thanks I will try that.

@Erftralle  It's very nice to see you here again

MrMusic:
Changes to fix the deprecated message with an empty string instead to remove the code.

Elfangor93:
Thank you very much!

PR #134, opened 2023-09-06
Update uppy uploader

Elfangor93:
Update the uppy uploader to the newest version.
This fixes the saving issue in the list of images inside the uppy view. Before this update only the last info entered into the TUS metadata fields was stored. Now this saving issue is fixed...


eumel1602:
Excellent. An important change! I have tested it. It all worked...

Issue #131, opened 2023-09-01
[Filesystem] Warning when upload new image


AlexanderSupp:
I tried NEW and MULTIPLE NEW and various parameters with ImageMagick.
The result was the same.
All images store correctly.
![Images-New-and-Multiple-New-Screenshot 2023-09-01 212915](https://github.com/JoomGalleryfriends/JG4-dev/assets/110494397/478ac685-5193-46b8-91e6-9341a16f8d7f)

**I think the reason is PHP 8.1.13. I changed back to 8.0.26 and everything works.**

A total multiple upload of 158 images shows also that this is the reason.
In this case 2 images were failed, but all images have the same warning.
![Screenshot 2023-09-01 215631](https://github.com/JoomGalleryfriends/JG4-dev/assets/110494397/48c8d5dd-002b-4fa6-b7ff-e5e96cc6ada2)
![Screenshot 2023-09-01 215649](https://github.com/JoomGalleryfriends/JG4-dev/assets/110494397/3eea6bde-109f-43e2-b5b1-927f4d83823b)

MrMusic:
This PR sould fix the Deprecated warning in the upload form: https://github.com/JoomGalleryfriends/JG4-dev/pull/133

AlexanderSupp:
I changed the PHP options to Joomla recommended values. I installed a fresh JG4 Alpha version. And I got no warning. From my point of view, this issue can be closed.

PR #136, opened 2023-09-06
Fix warning "Deprecated: Function strftime() is deprecated..." when editing images

MrMusic:
PHP 8.1: The use of the calendar field seems to generate this warning in Joomla in general.
It should be sufficient to change in the xml file to **_translateformat="true"_**. I don't know if there is a better solution?
See also: https://github.com/JoomGalleryfriends/JG4-dev/issues/131#issue-1877971439

### DIFF
--- a/administrator/com_joomgallery/forms/image.xml
+++ b/administrator/com_joomgallery/forms/image.xml
@@ -281,7 +281,7 @@
            format="%Y-%m-%d"
            timeformat="24"
            filter="server_utc"
-           translateformat="false"
+           translateformat="true"
            showtime="false"
            singleheader="false"
            todaybutton="true"


Elfangor93:
I propose not to change things here, but to wait until this issue is fixed in the core.
At the end of the day, it is just a warning and will not disturb users who doesn't have php error reporting to maximum.

The fix in the core is already detected:
https://github.com/joomla/joomla-cms/pull/41224 and https://github.com/joomla/joomla-cms/issues/41444

You may help the people there to fix this issue...

MrMusic:
>but to wait until this issue is fixed in the core.

I am not sure if this problem will be fixed by Joomla itself.
There are already several issues and pull requests on this topic at https://github.com/joomla/joomla-cms.
But it seems that the problem isn't solved in the long term.

>The fix in the core is already detected:
https://github.com/joomla/joomla-cms/pull/41224 and https://github.com/joomla/joomla-cms/issues/41444

In Joomla 5 this pull request was merged: https://github.com/joomla/joomla-cms/pull/41453
Wasn't the problem also solved here with `translateformat="true"`?
Someone should test whether the problem still exists in Joomla 5.
Let's wait and see.

PR #137, opened 2023-09-08
category view – css column grid & thumbnails

reilldesign:
This pull request contains code and style sheets to display a responsive css column grid with thumbnails. The structure is similar to the Joomla core and cassiopeia files.

 .com-joomgallery-category is a bit long as a class name. Currently there is a mix of joom and jg prefixes. I would suggest to use jg- uniformly.


media/com_joomgallery/css/site.css
```diff
@@ -1,3 +1,7 @@
+:root {
```
Elfangor93, line 1: grey-300 is already defined in template.css from cassiopeia. Do we still need this declaration?

media/com_joomgallery/css/site.css
```diff
@@ -59,6 +63,71 @@ table.itemList th.sort-cell .form-check-input {
...
+  }
+}
+
+@media (max-width: 768px) {
+  .com-joomgallery-category.images[class*=" columns-"] {
+    grid-template-columns: repeat(1, 1fr);
+  }
+}
+
+.com-joomgallery-category.image {
```
Elfangor93, line 100: This selector is empty...

Elfangor93 (changes requested):

Elfangor93:
>.com-joomgallery-category is a bit long as a class name. Currently there is a mix of joom and jg prefixes. I would suggest to use jg- uniformly

Agree.

eumel1602:
Hello. Thank you for your effort
It's also about displaying a category in FE?!
Attached is a picture of what it looks like on my test page. I don't know what your requirements are. Everything worked smoothly for me.
Since the image name is a link, the color of the font is "grey" and I would like a "bold" black letter better. I also find the title of the picture “in the middle” slightly better.

I don't know to what extent these PR settings are possible? I couldn't find anything ("square thumbnails", "access to image")
But that's probably not the point of PR.

Elfangor93:
Pull request closed since there is now a new one: See PR #138

PR #149, opened 2023-09-21
Update control panel

MrMusic:
Changes:
Fixes for Joomla 5
Show available extensions (JG3)
Control Panel design changes

`$this->addMailTemplate` does not work in Joomla 5.
@Elfangor93 Do you have any suggestions for this?


administrator/com_joomgallery/src/View/Control/HtmlView.php
```diff
@@ -156,7 +160,7 @@ protected function getMostViewedImages($imglimit)
 {
   $popularImages = array();
 
-  $db = Factory::getDbo();
+  $db = Factory::getContainer()->get('DatabaseDriver');
```
Elfangor93, line 163: This would be the correct way of retrieving the db class.
$db = Factory::getContainer()->get(DatabaseInterface::class);

grafik

Elfangor93 (changes requested):

Elfangor93:
Here I listed the most important deprecations for Joomla 4:
http://wiki.tech.spuur.ch/joomla/framework/deprecations

Elfangor93:
>$this->addMailTemplate does not work in Joomla 5.

I guess you refer to the addMailTemplate method in the script.php?
What exactly is the error message you get there? Because there is a Factory::getDbo(); in this method too. Maybe that's the problem?

script.php
```diff
@@ -427,10 +429,12 @@ function postflight($type, $parent)
 
       // Create default mail templates
       $suc_templates = true;
+/* auskommentiert wegen fehler in J5
```
Elfangor93, line 432: I don't like the approach to comment out code. This way the error message disappears and we will forget about the issue still beeing existent.
I prefer having an error than commenting stuff out.

Elfangor93 (changes requested):

administrator/com_joomgallery/src/View/Control/HtmlView.php
MrMusic: I had also already used this, but then followed the recommendations below:
https://forum.joomla.de/thread/19416-herstellen-einer-datenbankverbindung-innerhalb-eines-artikels-beitrags/?postID=140119#post140119
https://ghsvs.de/programmierer-schnipsel/joomla/334-jfactory-getdbo-ersatz-ab-joomla-4
Currently both solutions work

script.php
MrMusic: The function was only temporarily deactivated to be able to continue testing with an installable version and not to lose time.

MrMusic:
The error is
`0 class_exists(): Argument #1 ($class) must be of type string, Joomla\Component\Mails\Administrator\Table\TemplateTable given `
and occurs in line 461 of script.php.
The change to `$db = Factory::getContainer()->get(DatabaseInterface::class);` has no effect here.

Elfangor93:
>The error is `0 class_exists(): Argument #1 ($class) must be of type string, Joomla\Component\Mails\Administrator\Table\TemplateTable given` and occurs in line 461 of script.php.

Change the code to the following and the error is gone:
grafik

MrMusic:
@Elfangor93 Thank you. Error fixed.

administrator/com_joomgallery/src/View/Control/HtmlView.php
Elfangor93: Ja, funktionieren tut beides, zumindest bis und mit Joomla 6. Jedoch ist die Variante über das DatabaseInterface definitiv die bevorzugte Variante, da die Klassen unabhängig funktioniert. Das ist ja eben gerade der Vorteil von Interfaces...

MrMusic:
New error:
When editing an image, the following error occurs:
`0 Joomla\CMS\Form\Field\EditorField::filter field `imgtext` calls a deprecated filter class JComponentHelper, the class needs to be namespaced use \Joomla\CMS\Component\ComponentHelper instead or activate the backward compatible plugin. `

I have now changed the filter in the image, category and tag xml from
`filter="JComponentHelper::filterText"`
to
`filter="\Joomla\CMS\Component\ComponentHelper::filterText"`
I hope this is correct.

Elfangor93:
>I have now changed the filter in the image, category and tag xml from `filter="JComponentHelper::filterText"` to
`filter="\Joomla\CMS\Component\ComponentHelper::filterText"` I hope this is correct.

Yes, this seems to be correct. At least Joomla 5 is doing the same in the core ...
grafik

Elfangor93:
>After installing the JoomGallery, the 'start' and the 'languages' buttons appear. Is the 'languages' button still necessary?

No, I dont think so. I just copied that from JG3.

MrMusic:
For testing I have adapted the Quickicon plugin for Joomla 4 and put it up as available extension for download.
In the upper area of the control panel you could remove the white background, then it looks like this:
control-panel-transparent
Better or leave it white??

rowi68:
Upper area looks better now...

MrMusic:
Instead of displaying the defined image top lists above the news area, i set now a link to the image manager.
The image manager already provides all necessary functions to sort the images by any criteria. This saves a lot of code in the HtmlView.php.
The demand number of images to be displayed is no longer necessary.
The statistic and the info area are now a bit extended.
From my side the PR is now finished.
Are there still change requests?

MrMusic:
Last change: link text to be similar to the Joomla core

script.php
Elfangor93: Alright, lets do that:
Add "ToDo" to the comment. This way we can find lines of code again easily, which need to be fixed or reworked before the release.

MrMusic: >Add "ToDo" to the comment...

Yes, we can do that in the future. Here in this PR it is no longer necessary.

AlexanderSupp:
I installed this PR in my Joomla 5.0 JoomGallery 4 Test System with a concept for documentation.
http://joomgallery4.fotofreunde-much.de
Everything works fine, and the “Control Panel” looks good.
Before, I changed my Joomla 4.4 to 5.0 with a difficult process. 
The reason was, that my provider has not the recommanded database version. So I changed from MySQL to MariaDB.
See https://www.joomla.de/joomla/anforderungen

However, this PR is ready to merge into main as early as possible..
Thanks.

Elfangor93 (approved):

Issue #154, opened 2023-09-29
Categories, Status, Publish subcategory and parent category unpublished -A proper error message is expected.


AlexanderSupp:
Attempting to publish a subcategory whose parent category is unpublished will result in the error message COM_JOOMGALLERY_N_ITEMS_FAILED_PUBLISHING. A proper error message is expected.

Elfangor93:
This issue is resolved by PR https://github.com/JoomGalleryfriends/JG4-dev/pull/159.
Issue will be closed.

PR #159, opened 2023-10-02
Add missing language string

MrMusic:
Fix for issue https://github.com/JoomGalleryfriends/JG4-dev/issues/154

## Steps to reproduce the issue
Try to publish a category whose parent category is unpublished.

## Actual result:
Only the language key COM_JOOMGALLERY_N_ITEMS_FAILED_PUBLISHING appears.

## Expected result:
An error message that identifies the cause of the error.


MrMusic:
>I think we should use the same ones to be similar to the core.

Done.

PR #161, opened 2023-10-05
Fix error in script when upgrade from JG3 to JG4

Elfangor93:
Upgrade from JG3 to JG4 created an error, when script tries to uninstall modules during the upgrading process. Line 180 of script.php `$this->uninstallModules(array('mod_joomgithub'));` throws an error because it tries to call method `children()` on variable type array.

Thanks @csoscd for reporting this issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,5 @@
 /.idea/inspectionProfiles/Project_Default.xml
 /.idea/vcs.xml
 /administrator/com_joomgallery/joomgallery.xml
+/administrator/com_joomgallery/joomgallery_old.xml
 node_modules
--- a/administrator/com_joomgallery/src/Table/CategoryTable.php
+++ b/administrator/com_joomgallery/src/Table/CategoryTable.php
@@ -109,6 +109,18 @@ class CategoryTable extends Table implements VersionableTableInterface
 	}
 
   /**
+	 * Resets the root_id property to the default value: 0
+	 *
+	 * @return  void
+	 *
+	 * @since   4.0.0
+	 */
+  public static function resetRootId()
+  {
+    self::$root_id = 0;
+  }
+
+  /**
 	 * Define a namespaced asset name for inclusion in the #__assets table
 	 *
 	 * @return string The asset name
--- a/script.php
+++ b/script.php
@@ -59,11 +59,26 @@ class com_joomgalleryInstallerScript extends InstallerScript
   protected $new_code = '';
 
   /**
+   * Counter variable
+   *
+   * @var  int
+   */
+  protected $count = 0;
+
+  /**
    * True to skip output during install() method
    *
    * @var  bool
    */
   protected $installSkipMsg = false;
+
+  /**
+   * True to show that the current script is exectuted during an upgrade
+   * from an old JoomGallery version (JG 1-3)
+   *
+   * @var  bool
+   */
+  protected $fromOldJG = false;
 
 
 	/**
@@ -271,6 +286,7 @@ class com_joomgalleryInstallerScript extends InstallerScript
     {
       // We update from an old version (JG 1-3)
       $this->installSkipMsg = true;
+      $this->fromOldJG      = true;
       $this->install($parent);
     }
     else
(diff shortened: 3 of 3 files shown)


MrMusic:
I have tried the upgrate with this PR.
The JG3 tables were renamed correctly.
Unfortunately the default entries are missing in the new database tables.

OS: Windows
PHP 8.0.28
MariaDB 10.4.28

Elfangor93:
>Unfortunately the default entries are missing in the new database tables.

Yes, I have excluded this, because i didn't know how I want to solve this with the upgrade script. But now I know and we can include it.

MrMusic:
2 errors during installation:
`0 Resource 'DatabaseInterface' has not been registered with the container. `
and
`0 Resource 'LanguageFactoryInterface' has not been registered with the container. `

Reason:
`use \Joomla\Database\DatabaseInterface;`
and 
`use \Joomla\CMS\Language\LanguageFactoryInterface;`
is missing.

Elfangor93:
>Reason:
use \Joomla\Database\DatabaseInterface;
and
use \Joomla\CMS\Language\LanguageFactoryInterface;
is missing.

The two missing use stamenets are now in and the errror should be solved.

MrMusic:
>The two missing use stamenets are now in and the errror should be solved.

Thanks.
Sorry, another error still appears:
`0 Call to undefined method Joomla\CMS\Language\CachingLanguageFactory::getTag() `
in line 233 of script.php
Changing line 232 to 
`$language = Factory::getLanguage();`
solves the error for me.

Elfangor93:
`Factory::getLanguage()` is deprecated and will be removed in Joomla 6.
If the approach with the DI Container does not work we could try with `$app->getLanguage()`

grafik

MrMusic:
>If the approach with the DI Container does not work we could try with $app->getLanguage()

Yes, `$app->getLanguage()` works.

Now I make a test upgrade from Joomla 3.10.12 to Joomla 4.4.0
Procedure:
1. Successfully performed Joomla upgrade to 4.4.0.
2. Current branch 'script-uninstallModules' installed via the extension manager of Joomla.
3. Messages see screenshot
jg4-update-error-2023-10-18

4. After the installation attempt:
The folder 'com_joomgallery' in administrator/components is no longer exist
In the Database the old joomgallery tables are present with the extension '_old'
In the Database are no new tables from the JoomGallery 4 exist

5. Then second try to install the current 'script-uninstallModules' Branch
6. Messages see screenshot
jg4-update-error-2023-10-18_2

Installation now overall successful, but
In the Database the old tables now have 2x the extension "_old", e.g. "#__joomgallery_old_old"
and the new table #__categories does not contain root category.

AlexanderSupp:
Question:
When this PR fixed this error can I then directly upgrade to JG4 from Joomla 3.10 to Joomla 5.0, or must I proceed Joomla 3.10 to Joomla 4.4 and then from Joomla 4 to Joomla 5?

MrMusic:
>...can I then directly upgrade to JG4 from Joomla 3.10 to Joomla 5.0...

No, first you have to migrate to Joomla 4.4, See https://docs.joomla.org/Joomla_4.4.x_to_5.x_Planning_and_Upgrade_Step_by_Step

Elfangor93:
With Joomla 5 (or better PHP 8.2) there are still some issues everywhere in the code. Therefore up until now please use the JG4 only up until Joomla 4.4 and up until PHP 8.1.

Elfangor93:
Found the problem to this error:
276254968-9946a7e3-b64b-419a-9295-b3f83eae4de1

Since Joomla 4.4 and Joomla 5.x there is a change in the behavior of File::delete() depending if you use it as `Joomla\CMS\Filesystem\File::delete()` or `Joomla\Filesystem\File::delete()`. This is why in Joomla 5 and Joomla 4.4.x the installer script does not work anymore...
https://github.com/joomla/joomla-cms/pull/40257#issuecomment-1848936871

Elfangor93:
It has to be discussed how to solve this issue...

Elfangor93:
Now also this error is fixed
276255031-33878831-0710-4320-a10b-669b13d00000

@MrMusic Can you give this a last check?

MrMusic:
Error is fixed! After upgrade from Joomla 3.10.12/JoomGallery 3.6.2 this PR is installable and the default category was created correctly.

PR #164, opened 2023-10-26
Calculation of the rating

MrMusic:
This PR adds the functionality to calculate and display the rating of an image.
The calculate code was taken from JomGallery 3.6.
The rating is displayed in the backend and frontend at the respective image.
Please note: There is no functionality yet to rate images in the frontend. For testing, the number of ratings (imgvotes) and the rating sum (imgvotesum) must be entered manually into the database table.


rowi68:

AlexanderSupp:
Testing with Joomla 4.3.4, PHP 8.1.13, mysql 8.0.31.
The rating is limited to 5.
If I enter 1 / 6 the result is 1 / 5 and not 1 / 6.
If I enter 50 / 500 the result is 50 / 5 and not 50 / 10.
From my point of view, the rating must be limited by entering the data.
But not within the calculation process.

rowi68:
> The rating is limited to 5. If I enter 1 / 6 the result is 1 / 5 and not 1 / 6. 

I don't think this is relevant, as the evaluation will be done later on the website and not directly in the database. So it will not be possible to award a 6. 
In my opinion, everything is fine...

AlexanderSupp:
Keep in mind that JG 3 allows a rating up to 10.
In the config data, you can decide the limit for each installation.
More about this you find here: http://joomgallery4.fotofreunde-much.de/en/requirements/the-photo-club

rowi68:
Sorry, I didn't have that in mind. I assumed that 5 is the maximum rate.

MrMusic:
>If I enter 50 / 500 the result is 50 / 5 and not 50 / 10.

You are right. The value in the configuration is assumed to be the 'limit' for the display. However, this was already the case with JoomGallery 3.6.


>From my point of view, the rating must be limited by entering the data.

This is correct.
When submitting a rating, the maximum value specified in the configuration must not be exceeded.

>...as the evaluation will be done later on the website and not directly in the database. So it will not be possible to award a 6.

That is also correct too.
Therefore, there should be no a wrong rating calculation.

Elfangor93:
Can we merge this PR?
It has still a "needs testing" label...

AlexanderSupp:
This PR can merged. The max vote will be add later in the system.

Elfangor93:
@MrMusic Can you please resolve the remaining confict?
Then I guess we can merge...

MrMusic:
>Can we merge this PR?

Yes, or is there something against this?

>It has still a "needs testing" label...

Actually, every PR should be tested before it is merged.
The label is always removed after it has been merged, isn't it?

>@MrMusic Can you please resolve the remaining confict?

Should now be done.

Elfangor93:
Thank you very much ✅

Issue #162, opened 2023-10-13
Folders of categories are not deleted 


MrMusic:
When deleting categories, the folders remain in the filesystem.

### Steps to reproduce the issue
Create a category with title 'abc'.
->The category will be created correctly in the database and in the filesystem.

Try to delete the category 'abc'
->The category will be deleted in the database.
->The folders remain in the filesystem

@Elfangor93:
I suspect the error in the function '`getCatPath`' respectively in the '`$this->component->getFilesystem()->cleanPath`' of FileManager.php used there.
**Strange**: 
The function 'getCatPath' returns a corrupt path for **thumbnails**:
'/images/joomgallery/thumbnails\abc' becomes '\images\joomgallery humbnails\abc'.

But for originals and details the path is ok:
'\images\joomgallery\details\abc'
'\images\joomgallery\originals\abc'

### System:
OS: Windows
PHP 8.1
DB: 10.4.27-MariaDB

AlexanderSupp:
Joomla version 5.0. Provider Domain Factory, PHP 8.1, Maria-DB.
I tried to reproduce this issue. So I created a category “ABC”. I added an image. And then I tried to delete the image.
The result was:
![Screenshot 2023-10-30 190832](https://github.com/JoomGalleryfriends/JG4-dev/assets/110494397/b60fd83d-cce2-48e0-b2c1-3c84ca8fe37d)
But all 3 images were stored in the files correctly.
I tried it a second one with the same result.
All images are stored.
I repeat my test tomorrow.

AlexanderSupp:
I now can confirm the issue. I created a category and after than I deleted this category.
The result in Joomla 5 are below:
![Screenshot 2023-10-31 112338](https://github.com/JoomGalleryfriends/JG4-dev/assets/110494397/91683ab6-ad93-4e25-977d-c4d2c85fbe7a)

Elfangor93:
Issue is resolved by PR https://github.com/JoomGalleryfriends/JG4-dev/pull/166.
Close this issue.

PR #166, opened 2023-10-29
Fix delete empty category

MrMusic:
Possible Fix for issue https://github.com/JoomGalleryfriends/JG4-dev/issues/162

Before this fix: 
Deleting a empty category does not delete the folder in the filesystem.

After this fix: 
The folder in the filesystem shoud be deleted.


rowi68:
If deleting empty category, folder in the filesystem is also deleted...

AlexanderSupp:
See my test results in Joomla 5.0 documented in issue https://github.com/JoomGalleryfriends/JG4-dev/issues/162
When this issue here is solved, is this also in Joomla 5.0 solved?

MrMusic:
>When this issue here is solved, is this also in Joomla 5.0 solved?

Unfortunately, it looks to me as if Joomla 5 is adding further problems.
Example: Deleting categories under Joomla 5: In line 1103 of FileManager.php there is the following line: `if(\is_object($cat) && $cat instanceof \Joomla\CMS\Object\CMSObject && isset($cat->path))`.
The problem here is `$cat instanceof \Joomla\CMS\Object\CMSObject`. In Joomla 4.4 this was still valid but in Joomla 5.0 it is not anymore. I don't know why this is so and how to fix the problem. 
Simply removing `$cat instanceof \Joomla\CMS\Object\CMSObject` at this point would only be a temporary workaround. This code is used in many other places.
We'll have to wait what @Elfangor93 means about.

AlexanderSupp:
> We'll have to wait what @Elfangor93 means about.

Thank you. Hopefully Joomla 5 won't be a major construction site.

Elfangor93:
Yes, I know. CMSObject is removed with Joomla 5.
I can explain you that on the next meeting...

AlexanderSupp:
Testing with Joomla 4.3.4, PHP 8.1.13, mysql 8.0.31.
I can confirm the result from @rowi68.
PR works successfully and is ready to merge into main.

Elfangor93:
If I am informed correctly CMSObject is replaced with Registry. Maybe you could exchange the is instance of...

MrMusic:
>If I am informed correctly CMSObject is replaced with Registry. Maybe you could exchange the is instance of...

Simply replacing `instanceof \Joomla\CMS\Object\CMSObject` with `instanceof \Joomla\Registry\Registry` unfortunately does not fix the error.

MrMusic:
Changes:
- 'instanceof' removed
- function deleteCategory changed
- Removed conversion of the first letter to uppercase. Because: with some errors the name of the folder is displayed. The first letter is always converted to uppercase (ucfirst) although I had intentionally entered lowercase letters. To avoid such confusion I have removed this conversion in the FileManager.php

The following should now work in J4 and J5:
- delete a category - if empty
- error message if the category still contains images or subcategories
- There remains a possible 'ToDo': The possibility to delete categories that still have content (force_delete): But this would probably require major changes.

AlexanderSupp:
Uninstalled JG4 and installed this PR into Joomla 5.0
Screenshot 2023-11-14 183938

I built up all data from scratch.
I try to reproduce the issue #162 
errors are gone. Everything fine.
No problem with CMSObject.
I try to delete a category with content.
Response correct: Category-Folder with name 'wanderboys' could not be deleted because there are still files and subfolders in it.
I delete all images and repeat the catalogue delete. Everything fine.

It seems that this PR is ready to merge into main.
However, I continue my tests with this 5.0 installation.

rowi68:
I can confirm result from @AlexanderSupp.
Successfully tested with J4 and J5  :+1:

Elfangor93:
> 'instanceof' removed

I am not satisfied with this quick fix. It does not solve the problem but suppresses the negative effects. I will start working on a PR removing CMSObject from our extension anyway soon. With this the error will be properly fixed.

AlexanderSupp:
Is it possible to merge this PR now into main? And then close this PR, and we wait of a new PR.
There are too many open PR, with label "needs testing".

PR #172, opened 2024-01-01
Frontend category view

Elfangor93:
Extending the frontend category view.

grafik

`$this->item->parent`
Item object of the parent category. Information about the parent category.

`$this->item->children`
List of subcategory item objects. Information about the subcategories.

`$this->item->images`
List of image item objects of this category. Including pagination and filter objects to create a pagination and a filter form. 

### How to test
Visit `index.php?option=com_joomgallery&view=category` (category view) in the frontend. There is now a filter form on top of the image grid and a pagination below.
grafik

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@
 /administrator/com_joomgallery/joomgallery.xml
 /administrator/com_joomgallery/joomgallery_old.xml
 node_modules
+joomgallery_old.xml
--- a/administrator/com_joomgallery/layouts/joomla/form/field/radio/configbtns.php
+++ b/administrator/com_joomgallery/layouts/joomla/form/field/radio/configbtns.php
@@ -54,7 +54,10 @@ $classToggle = $isBtnGroup ? 'btn-check' : 'form-check-input';
 $btnClass    = $isBtnGroup ? 'btn btn-outline-secondary' : 'form-check-label';
 $blockStart  = $isBtnGroup ? '' : '<div class="form-check">';
 $blockEnd    = $isBtnGroup ? '' : '</div>';
-$configName  = explode('__', $id)[1];
+
+// Catch config name from form field name
+$configName_arr = explode('jg_', $name);
+$configName = preg_split('/[!"`#%&.,:;<>=@{}~\$\(\)\*\+\/\\\?\[\]\^\|]+/', end($configName_arr))[0];
 
 // Add the attributes of the fieldset in an array
 $containerClass = trim($class . ' radio' . ($readonly || $disabled ? ' disabled' : '') . ($readonly ? ' readonly' : ''));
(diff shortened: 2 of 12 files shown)


eumel1602:
I'm assuming this PR is just about the category view functionality?? (Images not sorted? I didn't discover any errors in the images at first)

When I click on the "Back to: Parent Category" button from a root category (main category), I get an error message "0":
Contents:
```
0
Please provide a valid record ID, alias or filename.
```

AlexanderSupp:
When the entered filter value does not match, I have no chance to continue. I expect a specific message and the possibility to continue. If the entered filter value is present, the images are displayed.

AlexanderSupp:
The number of possible pages is incorrect. With only 2 images and a page content of a possible 20 images, only one page is correct. If the last page is displayed, the next page always shows the images from the last page.
Screenshot 2024-01-02 192301

AlexanderSupp:
The allocation of the number of images per page works correctly. The sorting shows a result.
Perhaps the fixes to the issues mentioned here are not addressable in this PR. 
I can not say anything about this.

Elfangor93:
@AlexanderSupp The number of sites in the pagination should be fixed. But it irritates me that the search field is displayed for you but it shouldn't.
grafik

Can you please check, if the searchtools layout file is available in your installation and if there is no template override for that file.
grafik

MrMusic:
>Can you please check, if the searchtools layout file is available in your installation and if there is no template override for that file.

There is no layout folder in components/com_joomgallery...
I think the folder entry is missing in the joomgallery.xml: `<folder>layouts</folder>`

rowi68:
Works correctly.
I have noticed 2 things:
1. The categories are sorted in descending order and i see no way to set this ( or I don't know where )
2. Empty catgories are not displayed

Elfangor93:
>The categories are sorted in descending order and i see no way to set this ( or I don't know where )

Subcategories should be sorted as "Ordering ascending" and it is done with the code
`$listModel->setState('list.fullordering', 'a.lft ASC');`
Its not ment to be changed within this PR.

>Empty catgories are not displayed

This is correct. The following subcategories are not displayes: unpublished, hidden, empty
Therefore a subcategory must be published, not hidden and must have at least one image to be displayed.

I will not change this behavior in this PR since the PR focuses on pagination and ordering of images, not subcategories.

rowi68:
Have made new installation of this PR and "Ordering ascending" of categories is correct. :+1:

reilldesign:
The introduction of pagination and the option to select the number of photos displayed is a great feature and works perfectly in my test installation.

Please add the two values previous and next to the following link classes.

<a class="page-link **previous**" aria-label="Go to previous page" ...

<a class="page-link **next**" aria-label="Go to next page" ...

This will make it easier for me to integrate the Infinity scroll (load more button) in the next pull request.

PR #173, opened 2024-01-02
Fix wrong version info

MrMusic:
Under certain circumstances, the version info of the gallery is displayed incorrectly in the control panel.
This PR fixes this rare problem.

It also removes a duplicate entry from the joomgallery.xml:
`<plugin group="webservices" plugin="joomgallery"/>`

There's not much to test.


Elfangor93 (approved):
Looks good to me.

Issue #171, opened 2023-12-22
Frontend - Thoughts and discussion about the status at end March 2024 


AlexanderSupp:
Thoughts about the status and first official release in a nutshell

Joomla Version 5.0.1, 10.4.32-MariaDB, PHP Version 8.1.24
Uninstalling previous JG4 and install actual JG4 main from 22.12.2023
Preparation of Config Sets, Tags, Categories, and images from beginning. 
Everything works fine.

Status at the end of 2023:
Backend is ready for first release of JG4.
Frontend work in progress.

Target group of passive website visitors has priority.
Target group of active website visitors (upload and management of images) later.

Missing module JoomTreeView

The module shows the categories and subcategories of the gallery in the way you know from an explorer showing a filesystem. It gives a quick overview of the total existing categories in the gallery and is ideally suited to quickly and easily navigate between them.

That function is a great help for the first usable release.
Perhaps @erftralle can provide information on how this module can be migrated.

@reilldesign Category View working in progress.
Parameters (Global Value, values per category, values per user?):
Category Layout (Justified, Masonry, Grid Layout)
Title Alignment (None, Left, Center, Right)
Author Alignment (None, Left, Center, Right)
Recording Date (None, Left, Center, Right)
Ideal Height (in pixel)
Gap (in pixel)
Columns (1-6)
Image Layout (None, Boxed)
Sort Table by (...)
Count of Objects per page (5, 10, 15, 20, 25, 30, 50, ..)

Pagination

If I change the Count of Objects in "List of Images", this internal count will adapt for "Category View Justified". Until 25, 30 the images are well present at. Then the next higher is 50 and only max 2 images in one row are presented. But all 50 are displayed.
Grid Layout boxed are OK. Masonry also OK.

All movements within the navigation are carried out correctly. The layout parameters are used in the background and cannot be understood by the administrator or user.

I like @reilldesign idea of using parts of the LightGallery to display individual images in JG4.

I had no problems with Joomla 5.0

I'm looking forward to testing the first functions of migration.

AlexanderSupp:
Now another quarter of a year has passed.
For our next meeting, I suggest defining some of our front-end goals in a little more detail.
Parameters which we want to make available at which place to the administrator for the front end?
Level of Parameters (Global Value, values per category, values per user?):
Category Layout (Justified, Masonry, Grid Layout)
Title Alignment (None, Left, Center, Right)
Author Alignment (None, Left, Center, Right)
[...]

PR #174, opened 2024-01-04
Add Configuration options for the frontend

MrMusic:
First proposal for configuration options to set the display in the frontend.
There are options for the category view and for the detail view.
What does not work is the option 'Use global' in the menu item. This still needs to be implemented.
It is to be discussed which options should be changeable in the menu item at all.
It should also be discussed whether the layout should be separately adjustable for the display of the subcategories. Reason: The 'Justified' layout is very nice for displaying images. It is currently not so suitable for displaying subcategories. It would be nice if you could set Grid or Masonry for subcategories.


site/com_joomgallery/tmpl/category/default.php
```diff
@@ -161,9 +164,23 @@
           </div>
           <?php if ( $caption_align != 'none' && $category_class != 'justified') : ?>
           <div class="jg-image-caption <?php echo $caption_align; ?>">
+            <?php if ($this->config->get('jg_category_view_show_title')) : ?>
```
Elfangor93, line 167: I recommend to always set also the second argument whenever you use a get() method (which is the default value). This makes sure the default value is used, if the corresponding property/value is not available or not found.

Elfangor93:
>What does not work is the option 'Use global' in the menu item. This still needs to be implemented.

What is it that is not working?
In the backend category and image form the 'Use global' is successfully implemented. In order to make it work, you need to add the following layout to the form field:
`layout="joomla.form.field.radio.configbtns"`

If you need other fields with the 'Use global' option, then we will need to add a corresponding layout for that field...

MrMusic:
>In the backend category and image form the 'Use global' is successfully implemented. In order to make it work, you need to add the following layout to the form field of the menu item XML:
layout="joomla.form.field.radio.configbtns"

Thanks for the hint with the configbtns.php
I have now tried to display the field 'jg_category_view_show_title' in the menu entry as a simple test:
I added the layout="joomla.form.field.radio.configbtns" in the xml.
However, only the label is then displayed in the menu entry.
After I have copied the configbtns.php file from .../administrator/components/com_joomgallery/layouts/joomla/form/field/radio to /layouts/joomla/form/field/radio the field is displayed.
However, the field does not provide a correct result and show a warning message: `Warning: Undefined array key 1 in .../layouts/joomla/form/field/radio/configbtns.php on line 57`.
I suspect that the configuration cannot be read. Can't get any further here because I don't know where the error is. :-(

AlexanderSupp:
Needs discussion:?
See also my comments "Parameters (Global Value, values per category, values per user?):" in issue #171

Elfangor93:
@MrMusic Please try to use the official configbtns.php by adding `<folder>layouts</folder>` joomgallery.xml before installation. Does the warning also appears then?
grafik

MrMusic:
Adds further options to the configuration, these can be overwritten in the menu item if necessary.
This PR does not yet make any changes to the frontend. It is only used to discuss and define which settings should be made for the frontend in the configuration.
Currently, only options for the category view and the detail view are implemented.
Which options should be available in the configuration and be overwritable in the menu item?

Elfangor93:
Thank you very much...

PR #177, opened 2024-01-07
Add configuration dependent form fields

Elfangor93:
This PR adds form fields for configuration params that are inherited through the component global config sets, categories, images, menu items.

### The following field types are added:
- jglist (for selections)
  Dropdown list with first entry beeing "Use Global (value)"
  grafik

- jgradio (for yes, no selections)
  Three radio buttons with first button beeing "Use Global (value)"
  grafik

- jgtext (for free text input)
  Text field with hint beeing the global value
  grafik

- jgnumber (for number input)
  Number field with hint beeing the global value
 grafik

### Adding field to XML form

If you want the global inherited value to be calcultaed and displayed, its important to add the `useglobal="true"` attribute to the field definition in the XML.

grafik
Example of the jgradio field without predefined value for the global option. By default the global option value is an empty string "".

grafik
Example of the jgradio field with a predefined value for the global option. This might be useful if you want to show other fields conditionally by the global option value.

**Attention!**
If you want to use this fields in XML forms outside the joomgallery component, for example in a menu item form, its important that the `addfieldprefix="Joomgallery\Component\Joomgallery\Administrator\Field"` attribute is set in the form or fieldset you will use the field.

### How to test this PR
The parameter fields in the image and category form view should work the same as before.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@
 /administrator/com_joomgallery/joomgallery.xml
 /administrator/com_joomgallery/joomgallery_old.xml
 node_modules
+joomgallery_old.xml
--- a/administrator/com_joomgallery/forms/subform_catparams.xml
+++ b/administrator/com_joomgallery/forms/subform_catparams.xml
@@ -10,9 +10,9 @@
     <field name="jg_showrating"
            groupByFieldset="false"
            default="-1"
-           type="radio"
+           type="jgradio"
            class="btn-group btn-group-yesno"
-           layout="joomla.form.field.radio.configbtns"
+           useglobal="true"
            label="COM_JOOMGALLERY_CONFIG_ALLOW_RATING"
            description="COM_JOOMGALLERY_CONFIG_ALLOW_RATING_CAT_LONG">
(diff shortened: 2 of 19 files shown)


MrMusic:
I tried to include the fields in the menu item for the category view and to display the value from the configuration with Use global. It works quite well. The respective value is displayed 👍 
However, the following message always appears when opening the menu entry:
`Error: Configuration Set not properly loaded.`
The message is generated here in the DefaultConfig.php file:

    //---------Level 3---------
    if(isset($this->ids['category']) && $this->ids['category'] > 1)
    {
      // Load parent categories
      $cat_model = $this->component->getMVCFactory()->createModel('Category', 'administrator');
      $parents   = $cat_model->getParents($this->ids['category'], true);

      if($parents === false && empty($parents))
      {
        $this->app->enqueueMessage(Text::_('COM_JOOMGALLERY_SERVICE_ERROR_LOAD_CONFIG'), 'error');
        return;
      }

Elfangor93:
> However, the following message always appears when opening the menu entry:

Should be fixed now...

MrMusic:

Issue #101, opened 2023-05-20
[Image upload] Image record saved if error during imagetype creation


Elfangor93:
### Steps to reproduce the issue
1. Open an empty image form view (add new image)
2. Fill in the form and shoose an image
    --> Image should create an error during upload (e.g image >10MB, GD image processor, 128MB php max memory)
3. Press Save&Close

### Expected result
We stay in the view. An error stating that the item can not be saved and the debug output telling the problem.

### Actual result
View is closed. We return to the list view. An error appears that the image was not saved. Debug output is shown telling the problem. But still the image record is there, just without images.

### Additional comments
Imagetypes (original, detail, thumb) gets created after the image record is successfully stored to the database.
https://github.com/JoomGalleryfriends/JG4-dev/blob/94cd098ddab3a6b938a3160b4b7bef8533982e0a/administrator/com_joomgallery/src/Model/ImageModel.php#L492

If an error accured during imagetype creation, all images gets deleted (rollback) and the method is exited (return false).

Elfangor93:
Close this issue since PR https://github.com/JoomGalleryfriends/JG4-dev/pull/178 fixed this issue...

PR #178, opened 2024-01-11
Fix issue: Image record saved if error during imagetype creation

Elfangor93:
This is a fix for issue #101 .

New image records will not get saved anymore if there was an error during image creation.

### DIFF
--- a/administrator/com_joomgallery/src/Model/ImageModel.php
+++ b/administrator/com_joomgallery/src/Model/ImageModel.php
@@ -551,10 +551,13 @@ class ImageModel extends JoomAdminModel
 					$this->setError($this->component->getDebug(true));
           $uploader->rollback($table);
 
-          if($isNew || $isCopy)
+          if($isNew)
           {
-            // Delete already stored record since image creation failed
-
+            // Delete the already stored new record if image creation failed
+            if(!$table->delete($table->$key))
+            {
+              $this->component->setError($table->getError());
+            }
           }
 
 					return false;
--- a/administrator/com_joomgallery/src/Model/ImageModel.php
+++ b/administrator/com_joomgallery/src/Model/ImageModel.php
@@ -485,6 +485,18 @@ class ImageModel extends JoomAdminModel
       // Trigger the before save event.
 			$result = $app->triggerEvent($this->event_before_save, array($context, $table, $isNew, $data));
 
+      // Stop storing data if one of the plugins returns false
+			if(\in_array(false, $result, true))
+			{
+        if($imgUploaded)
+				{
+        	$uploader->rollback($table);
+				}
+				$this->setError($table->getError());
+
+				return false;
+			}
+
 			// Store the data.
 			if(!$table->store())
 			{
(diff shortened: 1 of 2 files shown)


Elfangor93:
> Whatever I do, Image Magick works and GD not. It shows me a blank screen.

@AlexanderSupp I am pretty sure this behavior is not related to the changes in this PR.

rowi68:
After uploading new image with an error during image creation (too big size) it will not get saved anymore.

But i noticed that I got an error with a picture size of 6MB although the limit is set to 10 MB.

2024-01-20_14h01_22

2024-01-20_13h58_16

If I set the maximum size on 15MB, then picture will accepted.

AlexanderSupp:
A second test performed from a scratch system.
Deinstalled JG. Checked DB-Tables, found xue7e_joomgallery_users_old_old_old, deleted this table.
Download JG4-dev-image-rollback at 20.01.2024 16:31
Installation OK.
Language German.
Upload without any changes of the defaults. Image size, 1402.86 KB / 1.33 MB, works correct.
Set max MB to 1
Upload same image, Error messages comes up, looks fine and is correct, because image too big.
Set max MB to 2
Upload same image, Error messages comes up, looks fine and is correct for me, because image size near a border.
Set max MB to 3.
Upload same image, No error messages, everything works fine.
Now I changed the parameter "Show warning" to "Yes, with hint". 
Max MB are still to 3.
Upload an image of 7 MB.
On a new, separate tab, the following occur:

Screenshot 2024-01-20 170617

I didn't remember exactly my parameters. I hope this test can reproduce the result from somebody.

MrMusic:

Elfangor93:
> But i noticed that I got an error with a picture size of 6MB although the limit is set to 10 MB.

Since we are using com_media's (Media Manager) filesystem plugins for the filesystem handling this error is coming from com_media and is not related to JoomGallery. Therefore this issue would need to be reported at the Joomla core.
https://github.com/joomla/joomla-cms

Elfangor93:
> I didn't remember exactly my parameters of the first test. I hope this test can reproduce the result from somebody.

@AlexanderSupp I am pretty sure, this issue is more general and will not be solved within this PR. Please open a separate issue for that.

rowi68:
> > But i noticed that I got an error with a picture size of 6MB although the limit is set to 10 MB.
> 
> @rowi68 Since we are using com_media's (Media Manager) filesystem plugins for the filesystem handling this error is coming from com_media and is not related to JoomGallery. Therefore this issue would need to be reported at the Joomla core. https://github.com/joomla/joomla-cms

I only noticed it and was not sure it would be interesting for you.  From my side it is not very important, because I usually never use pics of this size... :-)

AlexanderSupp:
> I only noticed it and was not sure it would be interesting for you. From my side it is not very important, because I usually never use pics of this size... :-)

Yes, that is my understanding too.
I'm testing PR here and I don't actually have any problems with it.
And in doing so, I run into further issues that I can no longer even assess. How should I open an issue that I can't properly define in the title?

MrMusic:
When creating an image, the wrong file size (default value from image.xml) is displayed, which is probably also causing confusion. Normally the value from com_media should be displayed here, analogue to the display in the configuration.
Otherwise I have not noticed any problems.

DE: Beim erstellen eines Bildes wird die falsche Dateigröße (Default-Wert aus der image.xml) angezeigt, vermutlich sorgt das auch für Verwirrung. Normalerweise sollte hier der Wert aus com_media stehen, analog zur Anzeige in der Konfiguration.
Ansonsten konnte ich keine Probleme feststellen.
new_image_filesize

Issue #157, opened 2023-09-30
[Category] Copy a category does not copy the corresponding image records


AlexanderSupp:
The marked categories are copied. A category with the addition “(0)” in the name is created at the same category level. The images in the physical file folders are also copied.
The number of images copied with it is not displayed in the category overview. The pictures are also not included in images. However, all images are available in the “details, originals, thumbnails” directories. Now, when I try to delete this orphaned category, I logically get the error message "Category folder named 'categoryname-0' could not be deleted because files and subfolders still exist."

Elfangor93:
When copy a category it successfully copies files and directories, but not the image record in the databse leading to this strange behavior.

AlexanderSupp:
What do we expect if we copy a category?
What happens with the subcategories? Are they copied too?
Make it sense to copy all images to the new categories? 
When will I use a copy function within categories?
Make it sense to duplicate the same images?
In my opinion, it is easier to leave out this copy function because the programming effort is disproportionate to the benefit.
Then we can also close this issue.

AlexanderSupp:
> DE: Ich würde erwarten dass nur der aktuelle Datensatz kopiert wird. Keine Bilder, keine Unterkategorien.

Dann gibt die Funktion (Batch, Duplicate) keinen Sinn, da ich dies auch mit einer Neuanlage erreichen kann.

MrMusic:
>Dann gibt die Funktion (Batch, Duplicate) keinen Sinn, da ich dies auch mit einer Neuanlage erreichen kann.

Nicht ganz: Bei Neuanlage sind alle Felder leer.
Beim Kopieren werden alle bereits vorh. Daten (Titel, Beschreibung, Params...) übernommen.

Elfangor93:
Close this issue since PR https://github.com/JoomGalleryfriends/JG4-dev/pull/179 fixed this issue...

Issue #153, opened 2023-09-29
[Category] Renaming a category


AlexanderSupp:
If I for example, create category “5” and now just save it without finishing the new creation, the 3 file directories are already created in the background. If I now change the category to “5 Buxtehude”, these directories will also be created. However, the directory “5” remains in the system. Such directories must be deleted manually.

MrMusic:
I can confirm the following problem:
If the alias of a category is changed manually, the 'path' in the database is not adjusted. Alias and path should be the same and unique.
Therefore, deleting the category after changing the alias is no longer possible.

Elfangor93:
**Here are two problems**
- When renaming a category, the folder is not renamed.
- When renaming a category, the path in the database is not adjusted.

AlexanderSupp:
Problem is solved with PR #179

AlexanderSupp:
The issue #157 is not solved. 
But with the changes in this PR the situation is cleared, if we drop the copy function of categories.
If so, we can close the issue 157 and this PR is ready to merge into main.

Elfangor93:
Close this issue since PR https://github.com/JoomGalleryfriends/JG4-dev/pull/179 fixed this issue...

PR #179, opened 2024-01-11
Fix issue: Wrong paths when renaming a category 

Elfangor93:
This is a fix for issue #153 .
When renaming a category (alias needs to change) the path is now correctly updated in the database.

This is also a fix for issue #157 .
When a category is copied, the folder content in the filesystem is no longer copied. Making it possible to perform further operations on the just created category through copying.

### DIFF
--- a/administrator/com_joomgallery/src/Model/CategoryModel.php
+++ b/administrator/com_joomgallery/src/Model/CategoryModel.php
@@ -464,7 +464,7 @@ class CategoryModel extends JoomAdminModel
           $source_id = $app->input->get('origin_id', false, 'INT');
 
           // Copy folder (including files and subfolders)
-          $manager->copyCategory($source_id, $table->path);
+          //$manager->copyCategory($source_id, $table);
         }
 
         // Clean the cache.
@@ -579,6 +579,7 @@ class CategoryModel extends JoomAdminModel
 	{
 		$app  = Factory::getApplication();
 		$user = Factory::getUser();
+    $task = $app->input->get('task');
 
 		// Access checks.
 		if(!$user->authorise('core.create', _JOOM_OPTION))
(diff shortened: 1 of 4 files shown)


AlexanderSupp:
Issue #153 is solved with this PR.
The issue https://github.com/JoomGalleryfriends/JG4-dev/issues/157 is not solved.
But with the changes in this PR the situation is cleared, if we drop the copy function of categories.
If so, we can close the issue 157 and this PR is ready to merge into main.

Elfangor93:
> But with the changes in this PR the situation is cleared, if we drop the copy function of categories.

With drop you mean hide it in the GUI?

AlexanderSupp:
Yes, that is correct. Instead of using the “Copy” function, we can simply add a new category, since a name “Category (2)” doesn’t make sense in most cases.

AlexanderSupp:
Oh, I see now it's a batch function duplicate! I mean the hole batch function too.

AlexanderSupp:
BTW: Save as Copy works fine.
But in some cases an error message "%s not found. Please provide one of the necessary info (object, id or alias)." appears.
Maybe it is when the parent category is root / - no parent -

Sorry, I'm a little confused with these two functions (Save as Copy and Batch Duplicate).

Elfangor93:
Save as copy and batch douplicate does exactly the same.

Batch douplicate just loops through the lost of selected items and performs the copy task for each of them.

Save as copy perfroms the copy task on the currently opened item.

AlexanderSupp:
@Elfangor93 
Yes, that was my last test result too.
Everything is going as planned, and therefore this PR can also be adopted in main.

However, there is a slight difference in the user's workflow.
Batch Duplicate creates an exact copy of the database entry with the changed name "Old name (2)".
In most practical cases, this entry is then touched upon again and changed.

With Save as Copy, I can make the changes straight away, which will then be taken into account in the new entry.
No further post-processing is necessary here.

Elfangor93:
First PR #178 has to be merged before this one...

PR #186, opened 2024-01-29
Add plugin: System - JoomGallery - Check owner

Elfangor93:
This plugin takes care of the connections between the different JG4 records against the #__users table of the Joomla core. It is the cleanest way to ensure the users in the `created_by` fields of the different records are and stay valid and available.

The plugin listens to the events `onMigrationBeforeSave`, `onContentBeforeSave` and `onUserBeforeDelete`. 

### onMigrationBeforeSave
The given user (owner) is checked for availablility and set to the fallback user if the provided owner does not exist. Event is triggered in the Migration-manager.

### onContentBeforeSave
The given user (owner) is checked for availablility and set to the fallback user if the provided owner does not exist. Event is called when a form edit view is saved (new & existing).

### onUserBeforeDelete
If a user gets deleted all record which are connected with this user are modified such that its owner is now the fallback user.


AlexanderSupp:
Joomla 5.0.2, PHP 8.1.13
Deinstalled JG4
Installed this PR.
Create categories. upload images, create 2 users.
Change owner in Image and category.
Delete user.
An error has occurred.
    0 Joomgallery\Plugin\System\Joomowner\Extension\JoomgalleryOwner::changeUser(): Argument #1 ($user) must be of type array, null given, called in C:\wamp64\www\Joomla5\plugins\system\joomowner\src\Extension\JoomgalleryOwner.php on line 257 
The user was not deleted.

rowi68:
I started with Joomla 4.4.2, PHP 8.1.27
Deinstalled JG4
Installed this PR.
Could not safe new categories or images because error:
2024-01-30_18h05_33

Then upgraded to Joomla 5.0.2
Could create categories, upload images, create users.
After delete user:
Es ist ein Fehler aufgetreten.
 0 Joomgallery\Plugin\System\Joomowner\Extension\JoomgalleryOwner::changeUser(): Argument #1 ($user) must be of type 
array, null given, called in /homepages/44/d260345546/htdocs/j4/plugins/system/joomowner/src/Extension/JoomgalleryOwner.php on line 257 
2024-01-30_18h29_20

AlexanderSupp:
Joomla 5.0.2, PHP 8.1.13
Installed main. Upload categories and images.
Installed this PR. The new fine English art, if I want to use old test data..
I assigned a new user as an owner to an existing image.
Then I deleted the user.
The owner of the image now is "None". That's fine.
The user is deleted. That's also fine.
Screenshot 2024-01-31 163036

Only the message failed.

reni68:
Joomla 4.4.2 PHP 8.1.17

1. User created
2. Installed this PR
3. created categories, 
4. upload images 
5. Owner  edited
6. this Owner deleted
7. => Fallback to configured owner

:+1:

Elfangor93:
Plugin funktioniert aktuell nur auf Joomla 4. Auf Joomla 5 gehts aktuell noch nicht...

rowi68:
I have tested with  Joomla 4.4.2 and 5.0.2 and got same results as @AlexanderSupp and @reni68

Elfangor93:
>Then I deleted the user.
The owner of the image now is "None". That's fine.

@AlexanderSupp After deleting the user which is owner, the owner of the image should be reassigned to the fallback user defined in the plugin settings.

AlexanderSupp:
If I set the owner to “No Owner” and save this image, an error occur.
An error has occurred.

    0 Joomgallery\Plugin\System\Joomowner\Extension\JoomgalleryOwner::isUserExists(): Argument #1 ($userId) must be of type int, string given, called in C:\wamp64\www\Joomla5\plugins\system\joomowner\src\Extension\JoomgalleryOwner.php on line 266

AlexanderSupp:
I got the same error by articles if I assign "No User".
danger
Save failed with the following error: Incorrect integer value: '' for column 'created_by' at row 1

So I assume that is a Joomla standard.

AlexanderSupp:
If I delete a user, the message looks like:
info
For the all JoomGallery images with IDs '1, 2, 182, 187', the old user ID '654' has been replaced with the fallback user ID '653'.
That's fine.
I can't assign "no user" in images and in the plugin settings. That's fine.
I found a user none in my images, which results from a failed upload of these images.

rowi68:
Tested successfully also with Joomla 5.0.2

I can confirm that with Joomla 5 it is not possible to assign "no user". But with Joomla 4.4.2 it works... :thinking:
Also the error "Save failed with the following error: Incorrect integer value: '' for column 'created_by' at row 1" seems to appear only with Joomla 5.

AlexanderSupp:
To complete:
In Joomla 3.10.12, article, "Kein Autor" possible.
In JG 3.6.2 category and image "Kein Benutzer" possible.

Elfangor93:
> If I set the owner to “No Owner” and save this image, an error occur.

Should be solved now. Can you please confirm?

Issue #181, opened 2024-01-13
[Configs] Change default value for image processing thumbnail


reilldesign:
All three available category views display landscape and portrait images well in combination. The masonry display only comes into its own when the image is reduced by width. It would be great to change the default setting to by width and 320px. Unfortunately I could not find the corresponding code for a pull request.

![Global-Configuration-Image-Processing-thumbnail](https://github.com/JoomGalleryfriends/JG4-dev/assets/43771555/d3744b46-4f4f-43fc-912c-b4cf1b587db7)

MrMusic:
>It would be great to change the default setting to by width and 320px.

Which is the optimal size is certainly a matter of taste.
For me personally, an ideal height of 200-250px would be better for the Justified layout. But it is certainly depends on the monitor size etc.
DE: Welches die optimale Größe ist, ist sicherlich Geschmacksache.
Für mich persönlich wäre beim Justified layout eine Ideal Height von 200-250px schöner. Ist aber sicherlich abhängig von der Monitorgröße usw.

>Unfortunately I could not find the corresponding code for a pull request.

The default values can be found in script.php. From about line 713 you will find the default values for the respective image type.
DE: Die Default-Werte sind in der script.php zu finden. Etwa ab Zeile 713 findest du für jeden Bild-Typ die Default-Werte (als JSON).

AlexanderSupp:
A resizing by width makes no sense to me. A resizing by height is good for panorama, landscape, and portrait. And here I prefer 250.
However, it's a matter of taste.

reilldesign:
The following pixel widths result for the standard Cassiopeia template within the component area with deactivated boxed view.

Masonry Layout

3 columns = w 418px
4 columns = w 305px 
5 columns = w 240px


Grid Layout

3 columns = w 320px
4 columns = w 305px 
5 columns = w 236px

A width of 250px produces a lower image quality in the vast majority of displays.

Since the 3 column layout is very common i suggest a value between 320px and 380px.

AlexanderSupp:
@reilldesign Thanks for this information. It's good to know something about the internals.

I tested some results of resizing the images.
Keep in mind that I resize not by width. I resize by height!

![grafik](https://github.com/JoomGalleryfriends/JG4-dev/assets/110494397/f02d0dc2-bc87-46eb-af8b-f69cc76b63ed)

As a consequence, I will change my resizing by Height to 380.

MrMusic:
[...]

PR #188, opened 2024-02-05
Image processing: Thumbnail by width

reilldesign:
Change the default setting of the thumbnail image processing to by width and increase the number of pixels to achieve good quality in a three-column layout.


Elfangor93:
To change the default image processing values, you need to adjust additionally the reset values in the `subform_imageconvert.xml`:
https://github.com/JoomGalleryfriends/JG4-dev/blob/05cca93c270c3b95cda3b9cde1b842042bd64c68/administrator/com_joomgallery/forms/subform_imageconvert.xml#L53

Elfangor93:
This PR is related to the discussion in issue https://github.com/JoomGalleryfriends/JG4-dev/issues/181

MrMusic:
After the 'Reset', the type of image resizing is still set to cropping, not by width.
I think line 40 of subform_imageconvert.xml needs to be adjusted from `reset="orig:0;det:3;thumb:4"` to `reset="orig:0;det:3;thumb:2"`.

PR #196, opened 2024-02-23
Error in JoomOwner Plugin

Elfangor93:
There was a bug in the `plg_system_joomowner` Plugin.

When saving a content type which does not have a connection to the users table, an error was thrown because the Plugin was not able to find the owner field in the table (since it does not exist in that case).

### How to test this PR

Before applying this PR, saving a configuration set ends in a PHP error.
After applying this PR saving a configuration set is successful.

### DIFF
--- a/plugins/system/joomowner/src/Extension/JoomgalleryOwner.php
+++ b/plugins/system/joomowner/src/Extension/JoomgalleryOwner.php
@@ -253,11 +253,11 @@ final class JoomgalleryOwner extends CMSPlugin implements SubscriberInterface
       return;
     }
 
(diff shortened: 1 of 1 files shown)


eumel1602:
Ich konnte keinen Fehler beim Speichern (verschiedene Test) der Configuration mehr finden.
J4.3.3 / PHP 8.1.19 / WIN

reni68:
Joomla 4.4.3 / PHP 8.1.17 / Linux

1. "main" installed
2. save configuration failed
3. this fix installed
4. save configuration successful

MrMusic:

PR #197, opened 2024-02-24
Fix php error when interrupt migration before step 1

Elfangor93:
If you interrupt a migration before saving the migration configuration form the first time you get a php error because there are no params.
This PR adds default values when getting migration params to mitigate that: `$this->params->get('same_db', 1)`

### Test result before applying this PR
Interrupting before submission of the migration configuration will throw a php error

### Test result after applying this PR
No error is thrown


eumel1602:
Auf zwei verschiednenen installationen getestet ! 
Es funktioniert nun gut! 
Unterbrechung der Migartion, ganz am anfang ohne was zu machen, funktioniert!
WIN / PHP 8.1 / J4.3.3

MrMusic:

PR #169, opened 2023-12-15
Show categories without images

MrMusic:
Currently, only (sub)categories that contain images are displayed in the category view.
This PR ensures that empty categories are also displayed.

PR #207, opened 2024-03-06
Show categories without images

MrMusic:
This is a redo of https://github.com/JoomGalleryfriends/JG4-dev/pull/169 but this has been cancelled (accidentally?).

Description: Currently, categories that only contain subcategories but no images are not displayed in the category view. 
But in my opinion, this should not be the case.
With this PR, empty subcategories are also displayed in the category view.

### DIFF
--- a/site/com_joomgallery/src/Model/CategoryModel.php
+++ b/site/com_joomgallery/src/Model/CategoryModel.php
@@ -424,7 +424,7 @@ class CategoryModel extends JoomItemModel
     $listModel->setState('filter.access', $user->getAuthorisedViewLevels());
     $listModel->setState('filter.published', 1);
     $listModel->setState('filter.showhidden', 0);
-    $listModel->setState('filter.showempty', 0);
+    $listModel->setState('filter.showempty', 1);
 
     if(Multilanguage::isEnabled())
     {


MrMusic:
Yes exactly, it is displayed in "List of Categories", but not in the **Category view**!
In your example, call up the 'Category view' of the category 'CAT-C'. The subcategory '111' should normally be displayed there, but currently it is not.
If you apply this PR it will be displayed.

DE: Ja genau, in "List of Categories" wird sie angezeigt, in der 'Category view' aber nicht!
In deinem Beispiel rufe die 'Category view' der Kategorie 'CAT-C' auf. Dort sollte normalerweise die Unterkategorie '111' angezeigt werden, derzeit wird sie das aber nicht.
Wenn du diesen PR anwendest wird sie angezeigt.

AlexanderSupp:
@MrMusic Thanks a lot. This problem is solved. 
I installed "JG4-dev-Show-categories-without-images" over an existing migrated installation and the results are fine.

But if I have a lot of subcategories, only the first "nn" are displayed. Where "nn" depends on the count of the image view which I can change in the image view. If I change these value, it will be not directly in effect for the categories. The effect will be transparent later.
This is what I pointed out in our last meeting.

Unfortunately, I have changed in my old 3.7.0 all categories, so I added an image if there was none. Also, I assigned to each category a thumbnail. I will test this PR again with Categories without images and thumbnails.

AlexanderSupp:
It works perfect without any images.
Screenshot 2024-03-07 130659

rowi68:

Before :

before_PR

After installing this PR:

with_PR

reni68:
> If you apply this PR it will be displayed.

I installed this PR over it and it's like you said. The subcategory is displayed even though there is no image.

eumel1602:
Ich habe diesen PR getestet. **Danke Alwin.** 
Es funktioniert bei mir mit Joomla 4.3.3 / Php 8.1.19 / WIN 10 / und Online

Elfangor93:
Thank you very much...

PR #58, opened 2022-12-22
Config service

Elfangor93:
## What is this service for

The Config service will be used to handle the different configurations set in the component. Based on the chosen inheritance method in the Glocal component settings, the correct set of configurations are loaded depending on the current context. The following configuration sets are taken into account:

- Global Configuration set (Config manager)
- User Configuartion set (Config manager)
- Category settings (Category edit form)
- Image settings (Image edit form)
- Menu item settings (Menu item edit form)

The following contexts are available:

- com_joomgallery (General settings)
- com_joomgallery.user.id (Settings for a specified user)
- com_joomgallery.category.id (Settings for all images in a specified category)
- com_joomgaller.image.id (Setting for a specified image)
- com_joomgallery.menu.id (Settings for a specified menu item)

## How does it work
When you instantiate the service, you have to provide a context and optionally an ID telling the service for which part of the component you want to have the configuration set loaded.
The service then will get the configuration sets and override the settings based on the chosen inheritance method. Default inheritance method will override the settings as follows:

1. Global Configuration set
2. User Configuartion set
3. Category settings (taking into account all parent categories)
4. Image settings
5. Menu item settings (frontend only, lookign which site are you accessing)

## How to use this service

You can load the service anywhere in Joomla by doing

```
$component = Factory::getApplication()->bootComponent('com_joomgallery');  // boot JoomGallery component
$component->createConfig('context', ID);   // instantiate service based on the provided context
$config = $component->getConfig();   // load the service into your scope
$config->get('jg_param'); // get parameter with key name 'jg_param'
```

## How to test the service

Make sure "Debug System" is "Yes" in Joomla Global configuration.
Open the file `/administrator/com_joomgallery/src/View/Faulties/HtmlView.php` and insert the following code on line 62 after `throw new \Exception(implode("\n", $errors)); }`

Different test instructions see comments below...

PR #211, opened 2024-03-18
Caching for config service

Elfangor93:
The config service was added with PR #58 .

The config service handles all the different configuration parameters and is used to get a parameter for a specific context. Since parameters are used everywhere within the component, the config service is called multiple times even within one php script runtime.
In the past, each time the service was called, a new instance was initiated and populated with the correct values for the individual parameters according to the chosen inheritance rules. The population of the correct values is quite time consuming. This PR adds a double layer caching mechanism to the config service in order to speed up this process.

### Static property caching
In PHP there exist the concept of the static properties for variables. If you declare an object property as static, the content of the property is available across different instances of an object and even survives the construction and deconstruction of instances. Therefore its perferctly suitable to realise a chache storage for object properties for the durage of the script runtime.

The static property `self::$cache` in the config service serves as a storage for prepopulated parameter values. For each context a simple unique string is created, representing the current combination of context and inheritance method. The unique string is created using the following scheme:
`ConfigName:context:usergroup:own`
This way parameters can be restored from cache instead of recalculating every time during one script runtime.
https://www.exakat.io/en/cache-with-php-static-variables/

### Session storage caching
To make precalculated parameter sets available over multiple script runtimes, the content `self::$cache` is stored to the session storage using `Factory::getSession()->set('com_joomgallery.configcache.ConfigName', self::$cache);`.
When a config service is created, the session store is copied back to static property.

To conclude.
This PR does not change the functionality of the config service but adds a caching mechanism to speed up the script runtime especially when the same set of inherited parameters are used multiple times within the same script runtime and even within the same session lifetime.

## How to test this PR
Click through everything. Everything should work exactly the same as before.
Additionally the testing procedure from PR #58 can be used. 

### Scenario 1: Usage of caching system

Put the following to any JoomGallery view (I use the faulty view for testing.)
```
$this->component->createConfig('com_joomgallery.user.id', 821);
$config = $this->component->getConfig();
dump($config)
```
Then you get the following if you browse to the view:
grafik

Important here is the `storeId`. This is the string used to define the configuration set. It should have the form `ConfigName:context:usergroup:own`

Where...
- ... ConfigName: Name of the config service in use (corresponds to the name property)
- ... context: The context of the content type. This corresponds to the first argument used when calling createConfig()
- ... usergroup: The usergroup of the current user or of the user selected if context is of type user. This should correspond to the usergroup selected in the user edit form under the tab "JoomGallery".
- ... own: 1 if the service is used with the own-property functionality (this is per default to true)

Add a Xdebug marker at the line 55 in the file `/administrator/components/com_joomgallery/src/Service/Config/DefaultConfig.php`
grafik

When the faulty view is called for the first time, the script should skip the code within the if brackets. Each subsequent time the faulty view is called, the code in the if blocks should be executed and therefore using the cached configuration values instead of recalculating them.

### Scenario 2: Handling cache entries related to a specific user

Put the following code to the faulty view:
```
$this->component->createConfig('com_joomgallery.user.id', <USERID>);
$config = $this->component->getConfig();
dump($config)
```
Call the faulty view with this code applied multiple times and change the userid every time to another existing user. This collects multiple cache entries for different usergroups of different users (different <USERID>).

Add a Xdebug marker at the line where you put `dump($config)` to the faulty view. You can now see the entries within the cache growing with each call of the view:
grafik

Now change the code to
```
$this->component->createConfig('com_joomgallery.user.id', 821);
$config = $this->component->getConfig();
$config->emptyCache('user.821');
dump($config)
```
This new line deletes all cache entries of the usergroups user with id 821 is in from the cache. You should see in Xdebug that some of the cache entries are gone.

### Scenario 3: Handling cache entries related to categories

Put the following code to the faulty view:
```
$this->component->createConfig('com_joomgallery.category.id', <CATID>);
$config = $this->component->getConfig();
dump($config)
```
Call the faulty view with this code applied multiple times and change the catid every time to another existing category. This collects multiple cache entries for different categories (different <CATID>).

Observe the growing of entries using Xdebug the same way as before.

Now change the code to
```
$this->component->createConfig('com_joomgallery.category.id', 2);
$config = $this->component->getConfig();
$config->emptyCache('category');
dump($config)
```
This new line deletes all cache entries related to categories from the cache. You should see in Xdebug that some of the cache entries are gone.


AlexanderSupp:
System Info:
PHP Built On 	Windows NT ALEXANDERSMIFCO 10.0 build 26080 (Windows 11) AMD64
Database Type 	mysql
Database Version 	8.0.31
Database Collation 	utf8mb4_0900_ai_ci
Database Connection Collation 	utf8mb4_0900_ai_ci
Database Connection Encryption 	None
Database Server Supports Connection Encryption 	Yes
PHP Version 	8.1.13
Web Server 	Apache/2.4.54 (Win64) PHP/8.1.13 mod_fcgid/2.3.10-dev
WebServer to PHP Interface 	apache2handler
Joomla! Version 	Joomla! 5.0.3 Stable [ Kuboresha ] 20-February-2024 16:00 GMT
Joomla Backward Compatibility Plugin 	Enabled (classes_aliases:"1", es5_assets:"1")
User Agent 	Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:123.0) Gecko/20100101 Firefox/123.0 

I uninstalled JG4 and started over with this PR.

My tests were with 8 categories and 513 images.
First, I noticed that ImageMagick works well.
I tested with Nmb Parallel Processes 8
The definition of categories and the upload of total of 506 images works successfully.
7 images failed for the following reasons.

I believe these reasons are not part of this PR as I remember these errors in previous other tests. But I am not sure.

Between the following two error images, 2 images were created successfully. All pictures before and after were also error-free.
Screenshot 2024-03-21 120101
Screenshot 2024-03-21 120652

The other types of errors were as follows:

Screenshot 2024-03-21 121553

AlexanderSupp:
> Between the following two error images, 2 images were created successfully. All pictures before and after were also error-free.

Sorry, the above information is not correct.
The screenshots are two independent similar errors. The first error case complains about a "Duplicate entry" of an image that was successfully uploaded 3 images before.
I therefore suspect that there is still a problem with "Nmb Parallel Processes" 8.

Elfangor93:
I updated the config service to add a custom field in the edit user form. With this field its now possible to select which usergroup should be selected when calculating the configuration set if a user is part of multiple usergroups.

grafik

Elfangor93:
In my first comment, I added the testing instructions for the PHP testing.
@szepty-ziemi Would you have time to perform these tests?

Elfangor93:
Thanks @MrMusic for reporting issues over email. The issues should be fixed now. Can you please check?

Elfangor93:
Thank you @MrMusic for testing. I guess this PR is now ready to merge.

PR #219, opened 2024-04-19
Update Error/Logging functionalities

Elfangor93:
This PR fixes and improves the error and logging functionities provided by the JoomGallery extension class.
The following methods are available:

```
// Add debug message
setDebug($txt, $new_line=true, $margin_top=false, $log=false, $name=null)

// Add warning message
setWarning($txt, $new_line=true, $margin_top=false, $log=false, $name=null)

// Add error message
setError($txt, $new_line=true, $margin_top=false, $log=true, $name=null)

// Add a message to a logging file
setLog($txt, $priority = 8, $name = null)
```

With the argument `$log` you can activate to add also a logging entry. So the message will be outputted as well as logged. With `$name` you can choose the filename of the logging file where to output. The default file is `com_joomgallery.log.php`. If you add a logging name the file will look like `com_joomgallery.name.log.php`.

### Scenario 1 (manually set an error)
Edit a HtmlView.php file of your choice. Add the following line:

```
$this->component->setError('Your error message');
```

Activate "Debug System" in the Joomla global configuration. Access the view you edited. 

You should get the error message as a bootstrap alert box
grafik

Additionally you get the message in the Joomla internal debugging console
grafik

And you should see a file in your Joomla log directory called `com_joomgallery.log.php` where you find the error message
grafik

### Scenario 2 (error during upload)
Upload two faulty images via the uppy uploader.
You should see a file in your Joomla log directory called `com_joomgallery.log.php` where you find the error message thrown from uploading the faulty images.

### Scenario 3 (error during migration)
Force an error during migration in step 3 of the migration manager.
You should see a file in your Joomla log directory called `com_joomgallery.log.php` where you find the error message thrown from uploading the faulty images.

### DIFF
--- a/administrator/com_joomgallery/src/Extension/MessageTrait.php
+++ b/administrator/com_joomgallery/src/Extension/MessageTrait.php
@@ -158,33 +158,6 @@ trait MessageTrait
   }
 
   /**
-   * Log a message
-   * 
-   * @param   string   $txt       The message for a new log entry.
-   * @param   integer  $priority  Message priority.
-   *
-   * @return  void
-   *
-   * @since   4.0.0
-  */
-  protected function addLog(string $txt, int $priority = 8, string $name = null)
-  {
-    if(\is_null($name) && \is_null($this->logName))
-    {
-      Log::add($txt, $priority, 'com_joomgallery');
-    }
-    else
-    {
-      if(\is_null($name))
-      {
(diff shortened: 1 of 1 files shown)


MrMusic:
Thank you very much.

Elfangor93:
I will merge this PR after #215 is merged.

PR #231, opened 2024-05-26
Choose different filesystem in categories

Elfangor93:
This PR adds the functionality to choose a different filesystem for each category.
This makes it possible to create categories whos images are stored on a different drive or server for this specific category which makes the whole filesystem functionality much more flexible and opens a lot of new usecases for endusers.

grafik

But there are some implications to be discussed and considered:
- Do we wanna have the possibility of choosing different filesystem for all categories or only the top level categories?
- What happens when changing the filesystem while there are images in the category?
- What happens when moving a category in or out a category with a different filesystem?

My Proposition:
- Different filesystems are only possible in top level categories.
- A different filesystem can only be defined if there are no images and no subcategories.
- Changing the filesystem of a category with images or subcategories or moving from or into a category with a different filesystem is not possible. These actions are considered a migration and needs to be done using a migration script.


AlexanderSupp:
File system is also a parameter in Configuration Sets, General Settings, Common Settings. 
The Configuration Sets depends on User Groups.
It hasn't been clear to me for a while how this is supposed to work. Does that even make sense? Who would use something like that? With all the dependencies and consequences, it becomes very complex and not very transparent.
Now you suggest using such an option for the category too. You name some consequences straight away.
Dealing with such solutions is beyond my imagination.

Elfangor93:
> It hasn't been clear to me for a while how this is supposed to work.
It works exactlx the same as in previous JoomGallery 3.x and before. You can define different sets of configuration for each user group. Meaning that you can create different user experiences for different user groups.

>  Now you suggest using such an option for the category too.
Yes, this will allow you to have images stored in different locations. Images from category 1 are stored locally, images from category 2 are stored in Dropbox and images from category 3 are stored on Amazon servers, etc.
It could open a whole bunch of new possibilities and expansion options. Of course for small image collections this is way to much, but imagine an image collection like Pixabay. There you need the possibility to spread the files accross different servers.

eumel1602:
Zwei Sachen zu Beginn. 
- Wenn ich von den möglichen Problemen lese, wäre ich dafür das so einfach wie möglich zu halten. Also 1. Kategorieebene. 
Kann es nicht einschätzen ob eine Erweiterung dieser Funktion auf z.b. Unterkategrien bei Bedarf oder Wunsch der user nachgereicht werden kann oder ob das jetzt entschieden werden muss ?! 

- die Installation des PR bringt in Joomla 5.1 wieder:  
 
bildschirmabgriff

Elfangor93:
> die Installation des PR bringt in Joomla 5.1 beim speichern einer Kategorie wieder
Ich habe den PR mit einem rebase auf den aktuellen Stand (wie main branch) gebracht. Somit sollte der Fehler behoben sein.

PR #239, opened 2024-06-09
Button visability in list views

Elfangor93:
This PR fixes the visability of the buttons in the list views (Issue #237 ).
I had to adjust the access service for that. So please check that the permissions are still correct.

PR #242, opened 2024-06-11
Simplifying usage of addLog()

Elfangor93:
Before this PR `$this->component->addLog(string $txt, $priority = 8, string $name = null)` the second argument needed to be a integer with a proper value for the type of message based on the constants defined in `Joomla\CMS\Log\Log`.
With this PR it is now possible to choose the type of message by providing a string with the proper name of the message type.

**Before this PR**
`$this->component->addLog('My error message', 8, 'jerror')`
which is the same as
`$this->component->addLog('My error message', Log::ERROR, 'jerror')`

**After this PR**
`$this->component->addLog('My error message', 'error', 'jerror')`

Where the old way is still possible.

This fixes the issue reported in PR #239 

>0 Class “Joomgallery\Component\Joomgallery\Administrator\Model\Log” not found
1 () JROOT\administrator\components\com_joomgallery\src\Model\CategoryModel.php:297
2 Joomgallery\Component\Joomgallery\Administrator\Model\CategoryModel->delete() JROOT\libraries\src\MVC\Controller\AdminController.php:143
3 Joomla\CMS\MVC\Controller\AdminController->delete() JROOT\libraries\src\MVC\Controller\BaseController.php:693 

### How to test this PR
1. Open the file /com_joomgallery/administrator/src/View/Faulties and add the following code after line 49 (in front of the return statement):
`$this->component->addLog('My error message for testing.', 'warning', 'jerror');`

2. Visit the site with the following URL in your browser:
`http://<Your-Domain>/administrator/index.php?option=com_joomgallery&view=faulties`

3. Check that the message was added to the jerror log file
`/administrator/logs/com_joomgallery.jerror.log.php`

Repeat the test by changing the second argument in addLog() in step 1. Possible values for the second argument are:
all, emergency, alert, critical, error, warning, notice, info, debug

### DIFF
--- a/administrator/com_joomgallery/src/Extension/MessageTrait.php
+++ b/administrator/com_joomgallery/src/Extension/MessageTrait.php
@@ -177,15 +177,22 @@ trait MessageTrait
    * Log a message
    * 
    * @param   string   $txt       The message for a new log entry.
-   * @param   integer  $priority  Message priority.
+   * @param   mixed    $priority  Message priority.
    *
    * @return  void
    *
    * @since   4.0.0
   */
-  public function addLog(string $txt, int $priority = 8, string $name = null)
+  public function addLog(string $txt, $priority = 8, string $name = null)
   {
     $this->addLogger($name);
+
+    if(\is_string($priority))
+    {
+      $priority = \strtoupper($priority);
+      $constant = "Joomla\CMS\Log\Log::$priority";
+      $priority = \constant($constant);
+    }
     
     if(\is_null($name) && \is_null($this->logName))
     {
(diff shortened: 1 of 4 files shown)


MrMusic:
I have done some tests with the second parameter:
With numerical values (1, 2, 4,...) it works fine.
With strings ('error', 'warning', 'notice'...) it works fine.
What does not work is using the Joomla constants (Log::alert, Log::ERROR, ...). Would it be useful if you could also use the Joomla constants?

Elfangor93:
> What does not work is using the Joomla constants (Log::alert, Log::ERROR, ...).

These are not Joomla constants. They are constants defined in the Log class. In order to use that you need to load the Log class into your current scope by doing `use Joomla/CMS/Log/Log`.

MrMusic:
>In order to use that you need to load the Log class into your current scope by doing use Joomla\CMS\Log\Log.

That is of course, I have already added that.

The problem was actually that I didn't pay attention to upper/lower case when testing.
Log::WARNING -> ok
Log::warning -> error

MrMusic:
Thank you very much.

PR #258, opened 2024-09-05
Fix error in migration manager

Elfangor93:
In the migration manager the PHP error `Return value must be of type array` is thrown when trying to continue a migration. Additionally the ACL is done in the wrong way and also there an error is thrown.

This PR fixes this issues in the migration manager.

### DIFF
--- a/administrator/com_joomgallery/src/Model/MigrationModel.php
+++ b/administrator/com_joomgallery/src/Model/MigrationModel.php
@@ -1368,4 +1368,32 @@ class MigrationModel extends JoomAdminModel
       return $key;
     }
   }
+
+  /**
+   * Method to test whether a record can be deleted.
+   *
+   * @param   object  $record  A record object.
+   *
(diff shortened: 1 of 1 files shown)


MrMusic:
Thank you very much!

reilldesign:
I get this error message

Step 2: Migration pre-check failed. Failed check: Source extension compatibility

I found the following thread on this topic
[https://www.forum.joomgalleryfriends.net/forum/index.php?thread/403-joomgallery-categories-root-gel%C3%B6scht/&postID=2271&highlight=Source%2Bextension%2Bcompatibility#post2271](url)

which xml file must be located where? Is there any other way to fix this error?

reilldesign:
The J3 joomgallery_old.xml file was accidentally deleted due to the many de- and installations. After adding it manually, everything worked fine without an error message.


Thank you very much!

rowi68:
I have also tested this PR ✅ successfully.

Elfangor93:
Thank you all for testing!

PR #265, opened 2024-09-30
[Frontend] Router

Elfangor93:
Adding the router to the frontend. Making the URL's as close as possible to the ones of the core (com_content, com_users, ...). But having the possibility to switch to the old JG3 router for compatibility.

The new, modern Router (RouterView, since Joomla 3.8) was implemented. This changes quite a lot in contrast to the legacy router of the old JoomGallery. In order for the Router to work correctly, JoomGallery will create three Menu Items by default during installation. The main view where all the routing is attached to is the gallery view. The gallery view is implemented only very rough in this PR. Maybe we could add a nicer template to this view after merging this PR?

## How the router works
The modern router creates URLs based on found menuitems and creates the URL segments based on the hirarchy of the found menuitems. If there are any query parameters left which can not be translated with the available menuitems then they will be attached at the end of the SEF URL.

**Available menuitem types**
- Gallery view (root view)
- List of categories (root view)
- Category view (child of gallery view)
- List of images (child of gallery view)
- Image view (child of list of images)

**Example:**
`index.php/gallery/images/1-testimage.html?catid=1`
|-> `gallery` is the alias of a menuitem of type gallery
|-> `images` is the alias of a menuitem of type list of images
|-> `1-testimage` is the ID and alias of the image to be displayed
|-> `html` is the format to be displayed. Html means that the detail view of the image is displayed.
|-> `catid=1` there was a query parameter `catid` which could not be translated to a segment. This query parameter will be added at the end of the SEF URL.

## How to test this PR
Activate the SEF routing in the Joomla global configuration. Add content to the component and create different menuitems. Then flip through the component views in the frontend and check if
- No 404 (view not found) appears
- URL looks as if you would expect them knowing the core URL behavior and the description provided above


MrMusic:
Should the option 'Remove IDs from URLs' in the Component Options work?
If activated, it is not possible to navigate to other (sub-)categories.

Elfangor93:
Yes, it should work. 🤭

Elfangor93:
> Should the 'Breadcrumbs' already work too?

No. It will follow in another PR.

Elfangor93:
> The Detail view (Single Image View) still does not show a nice URL. See Test Page: https://mrmusic63.de/jg4-dev/alias-menue-category-view-root?view=image&id=1

Is there a menuitem of type `images` available in your installation?
This is needed for the modern router to produce correct urls.
If you install this PR the install script should create such a menuitem.

MrMusic:
>Is there a menuitem of type list of images available in your installation?
This is needed for the modern router to produce the urls nicely.

Thanks for the explanation.
I have now restored the menu entries on the linked test page.
So it is a very important thing that the invisible menu entries should not be deleted!

If there is no direct menu entry, the URL for the detail view looks like this .../alias-menue-gallery-view/alias-menue-list-of-images/cocktail-8308286-1920-jpg
If there is a menu link to a single image, the URL only contains the menu alias, not the image alias. Example: https://mrmusic63.de/jg4-dev/alias-menue-detail-view-ferris-wheel
Is this correct?

Elfangor93:
Yes, this is the correct behavior of the Joomla core router.

MrMusic:
>Yes, this is the correct behavior of the Joomla core router.

Ok.

I can't see any other problems.
Thank you very much.

PR #274, opened 2024-10-24
Change of the 'Image types' names

Elfangor93:
This PR fixes the issue reported in #213 
When compatibility mode activated, the old image type names gets forewarded to the new ones. This way the old image source paths should work again.

**Old (JG3)**
index.php?option=com_joomgallery&view=image&format=raw&id=123&type=img

**New (JG4)**
index.php?option=com_joomgallery&view=image&format=raw&id=123&type=detail

### DIFF
--- a/administrator/com_joomgallery/src/Helper/JoomHelper.php
+++ b/administrator/com_joomgallery/src/Helper/JoomHelper.php
@@ -402,6 +402,28 @@ class JoomHelper
     // Create file config service based on current user
 		$config = self::getService('Config');
 
+    // Adjust type when in compatibility mode
+    if($config->get('jg_compatibility_mode', 0))
+    {
+      switch($type)
+      {
+        case 'thumb':
+          $type = 'thumbnail';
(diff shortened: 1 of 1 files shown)


MrMusic:
>When compatibility mode activated, the old image type names gets forewarded to the new ones. This way the old image source paths should work again.

That works very well - thank you very much.
Perhaps it would make sense that this could/should also work without activated 'compatibility mode', i.e. simply without the `if` condition? Then you could migrate to the new file structure and the old code in the Joomla articles would still work.

Elfangor93:
I did that not to create issues when an imagetype was created with name "img". If this is the case, this imagetype would never get displayed but the "detail" imagetype instead.
So it will create strange behavior when imagetypes were created with exactly the names from the old types. This is unlikely but not impossible. Thats why I would leave it like that.

MrMusic:
>Thats why I would leave it like that.

OK.
Thank you very much.

PR #280, opened 2024-11-09
Router selection

Elfangor93:
This PR is building on top of PR #277.

This PR adds a new configuration parameter `Routing` where you can choose between different routing methods. The parameter `Remove IDs from URLs` was moved from the component parameters (index.php?option=com_config&view=component&component=com_joomgallery) to the JG config manager (index.php?option=com_joomgallery&view=configs).

This PR adds the possibility to add different routers to JoomGallery. Pre-Installed are two routers:
1. Modern Router (default). This is the router from PR #277.
2. Modern Router (JG3 flavor). This router creates routes similar to the ones from JG3.

---
grafik


MrMusic:
>The parameter Remove IDs from URLs was moved from the component parameters

Removing the IDs from the URL only applies to Categories, not to Images. Perhaps this should be noted in the description.

If you click on a parent Category in the breadcrumbs, an error occurs:
```
TypeError:
Joomla\CMS\Error\AbstractRenderer::getRenderer(): Argument #1 ($type) must be of type string, null given, called in ...libraries\src\Exception\ExceptionHandler.php on line 116

  at ...libraries\src\Error\AbstractRenderer.php:70
  at Joomla\CMS\Error\AbstractRenderer::getRenderer(null)
     (...libraries\src\Exception\ExceptionHandler.php:116)
  at Joomla\CMS\Exception\ExceptionHandler::render(object(Exception))
     (...libraries\src\Exception\ExceptionHandler.php:72)
  at Joomla\CMS\Exception\ExceptionHandler::handleException(object(Exception))
     (...libraries\src\Application\CMSApplication.php:322)
```
No matter which router is selected. 
See also: https://mrmusic63.de/jg4-dev/

Elfangor93:
> Removing the IDs from the URL only applies to Categories, not to Images. Perhaps this should be noted in the description.

But it really depends on the selected router. One could also add a router which does remove it also for images.

Elfangor93:
> If you click on a parent Category in the breadcrumbs, an error occurs

The error should be fixed.

Elfangor93:
Hätte jemand Zeit, diesen PR zu testen?

eumel1602:
frühestens am mittwoch. sorry.

MrMusic:
>The error should be fixed.

Danke für den Fix.

Die URLs sehen für "Modern Router (default)" und auch für Modern "Router (JG3 flavour)" gut aus.

In der Category view sehen die Breadcrumbs auch gut aus.

Nur die Breadcrumbs eines einzelnen Bildes sehen bei aktiviertem JG3-Router anders aus:
Original JG3 wäre z.B.: `JoomGallery >> Topkategorie >> Pixabay >> bridge-8198882_1920.jpg`
Modern Router (JG3 flavour): `JoomGallery >> Categories >> Images >> bridge-8198882_1920.jpg`

>One could also add a new router which does remove all IDs also for images. Just the two currently available are both not removing it.

Isn't the assumption that someone will add another router a bit optimistic? ;-)

Elfangor93:
> Nur die Breadcrumbs eines einzelnen Bildes sehen bei aktiviertem JG3-Router anders aus

Breadcrumbs haben nichts mit dem Router zu tun. Wenns nicht zusammen passt und es stört, kann man die Breadcrumbs ja deaktivieren.

MrMusic:
Thank you very much!

PR #293, opened 2024-12-05
Different issues during installation

Elfangor93:
This PR fixes the issue #291 and some other smaller issues during installation.

When installing the JG4 component multiple times while the old JG3 tables are still present, the tables will be renamed just once.


MrMusic:
Thank you very much.

Issue #298, opened 2024-12-09
Migration Manager error


hansdampf-123:
Migration fails at Step 3, Migration: Categories with the Error :
[Success] [Migrator.js] Migration of category with id = 1 successful.
[Info] Error in server response. We will try again. (1/3)
**[Fehler] Call to undefined method Joomla\Event\Event::getItem()**

<img width="1399" alt="Bildschirmfoto 2024-12-09 um 16 24 43" src="https://github.com/user-attachments/assets/29491a36-4aa0-4a90-aa2e-f16feb29b335">

source:
Joomgallery 3.6.2
Joomla 3.10.12

destination:
Joomgallery 4.0.0 dev
Joomla 5.2.2

I've done this kind of migration on a other side with out any problems, but here something is not working. Any idea?

Elfangor93:
This error was probably introduced with PR #292. Can you provide me with a full call stack of the error?
You can get that when looking into the network tab of your browser console...

MrMusic:
>[Fehler] Call to undefined method Joomla\Event\Event::getItem()

Den Fehler kann ich bestätigen:
```
[Error] Call to undefined method Joomla\Event\Event::getItem()

[Error] 
#0 C:\xampp81_12\htdocs\j5-test\libraries\vendor\joomla\event\src\Dispatcher.php(454): Joomgallery\Plugin\System\Joomowner\Extension\JoomgalleryOwner->onMigrationBeforeSave(Object(Joomla\Event\Event)) 
#1 C:\xampp81_12\htdocs\j5-test\administrator\components\com_joomgallery\src\Model\MigrationModel.php(1290): Joomla\Event\Dispatcher->dispatch('onMigrationBefo...', Object(Joomla\Event\Event)) 
#2 C:\xampp81_12\htdocs\j5-test\administrator\components\com_joomgallery\src\Model\MigrationModel.php(807): Joomgallery\Component\Joomgallery\Administrator\Model\MigrationModel->insertRecord('category', Array, false) 
#3 C:\xampp81_12\htdocs\j5-test\administrator\components\com_joomgallery\src\Controller\MigrationController.php(809): Joomgallery\Component\Joomgallery\Administrator\Model\MigrationModel->migrate('category', 28)
```

hansdampf-123:
After upgrade (including #283) at the same step. These Errors show up:
(Joomla 5.2.2., Joomgallery deleted and newly installed)

[...]

PR #301, opened 2024-12-14
Fix issue with onMigrationBeforeSave event

Elfangor93:
This PR fixes the issue reported in #298 

The way how the onMigrationBeforeSave event is triggered has to be changed now after PR #292 is merged. In Joomla 5+, we have to use the dedicated events like `Joomla\CMS\Event\Model\BeforeSaveEvent` containing special methods like `getItem()`. The `Joomla\Event\Event` method can not be used anymore in Joomla 5+ while it still has to be used in Joomla 4.x.

In order to support both Joomla versions, we need a switch and both ways of triggering the event. Thanks @joomla/joomla-cms for this additional work ;-)

--> Here you can see now the double work to be done by developers if you want to have backward compatibility!

### How to test this PR
The migrations should be working again with this PR applied.

### DIFF
--- a/administrator/com_joomgallery/src/Model/MigrationModel.php
+++ b/administrator/com_joomgallery/src/Model/MigrationModel.php
@@ -17,6 +17,7 @@ use \Joomla\CMS\Factory;
 use \Joomla\CMS\Log\Log;
 use \Joomla\CMS\Uri\Uri;
 use \Joomla\CMS\Form\Form;
+use \Joomla\CMS\Event\Model;
 use \Joomla\CMS\Language\Text;
 use \Joomla\Registry\Registry;
 use \Joomla\CMS\Filesystem\Path;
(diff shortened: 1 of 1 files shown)


rowi68:
Ich habe den PR ✅ erfolgreich getestet. Migration läuft problemlos durch. ( J5.2.2 PHP 8.2 )

Bei eingeschaltetem "Fehler berichten: Einfach/Maximum" erscheint jedoch über den Bildern bzw. Kategorien folgende (mehrere) Meldung(en):

Warning: Attempt to read property "owner" on bool in /...xxx/administrator/components/com_joomgallery/src/Service/Access/Base/AccessOwn.php on line 144

Elfangor93:
> Warning: Attempt to read property "owner" on bool in /...xxx/administrator/components/com_joomgallery/src/Service/Access/Base/AccessOwn.php on line 144

Bitte separates issue eröffnen. Wird nicht in diesem PR gefixt.

Issue #285, opened 2024-11-18
Error when deleting a static image type


MrMusic:
In the JoomGallery configuration you can define your own static image types.
If you want to delete such a image type, the following error occurs:
```
An error has occurred.

1 Invalid asset provided for ACL access check 

Call Stack
# 	Function 	Location
1 	() 	JROOT\administrator\components\com_joomgallery\src\Service\Access\Access.php:480
2 	Joomgallery\Component\Joomgallery\Administrator\Service\Access\Access->prepareAsset() 	JROOT\administrator\components\com_joomgallery\src\Service\Access\Access.php:166
3 	Joomgallery\Component\Joomgallery\Administrator\Service\Access\Access->checkACL() 	JROOT\administrator\components\com_joomgallery\src\Model\JoomAdminModel.php:676
4 	Joomgallery\Component\Joomgallery\Administrator\Model\JoomAdminModel->canDelete() 	JROOT\libraries\src\MVC\Model\AdminModel.php:843 
```
Joomla 5.2.1
PHP 8.1

Issue #252, opened 2024-06-30
Huge image collection slows down the whole Joomla


Elfangor93:
Currently JoomGallery creates an asset record (#__assets) for each image.

This allows us to have a permissions tab in the image edit form which which we can set permissions on a image level. This way we can e.g deny the edit of a specific image for one usergroup even if this usergroup is allowed to edit images in this category.
![grafik](https://github.com/JoomGalleryfriends/JG4-dev/assets/39154009/75511ada-6c02-4c00-8baf-4a3cf92e4156)

Lately we have seen that it slows down the asset table if we have a huge number of images (>20'000 images). With so many records in the asset table, each queries becomes very slow. This results in a low performance of the whole CMS because everything is using the asset table.

The only way to fix this issue would be to not store assets for images & tags. This would mean another huge B/C break, since JG3 stored assets for images. But it would also speeds up the performance of the whole website.

What do you think?
Do we need ACL on image level?
Do we wanna create a configuration to turn the creation of assets on or off?

AlexanderSupp:
I see no reason to define different permissions for individual images in a category. The regulations at the category level are completely sufficient. In my opinion, it also makes no sense to allow permission for different regulations in the configuration.
The "Own-" options can be used to regulate finer details.
Keep it stupid and simple.

Elfangor93:
Another idea would be to create just one asset per category to handle permissions for all the images in the category.
This way we are able to select the permissions for the images independently of the permissions of the categories.

MrMusic:
>Another idea would be to create just one asset per category to handle permissions for all the images in the category.

Would the permissions tab then appear for every image?
I guess this would confuse the users because they would think the setting would be separate for each image.

>What do you think?
Do we need ACL on image level?

I don't think we need ACL on image level. I currently have no 'use case' for this.

>Do we wanna create a configuration to turn the creation of assets on or off?

Would this not make the code more complicated and slow down the performance?

Tazzios:
And only create an asset record for the image if image authorization is NOT the same as the category/ an override has been set for the image.

I also do not see an use case for image ACL. For some special occasion maybe a few images, but if you use this option a lot i think you will lose oversight about who can see what image.

PR #309, opened 2024-12-21
Optimise asset entries for Collections, Tags, Comments, Votes, Imagetypes

Elfangor93:
This PR will add asset entries and therefore allow ACL checks for collections, votes and comments.
It introduces a third type of asset handling: Global handled assets

- Categories (Default assets)
- Images (Parent handled assets)
- Collections (Default assets)
- Configs (Default assets)
- Imagetypes (Global asset)
- Tags (Global asset)
- Votes (Global asset)
- Comments (Global asset)

So to summarize; JoomGallery now has three differt ways of creating assets and therefore to handle the acl for its content types. The following exist:

- **Default assets**
   This is the Joomla! core way of creating assets. Each new item gets one dedicated asset. This way the permissions for each item can bet set or overriten individually.
- **Parent handled assets**
   There is one asset created for all items belonging to the same parent item. Permissions can be set in the edit form of the parent item and it affects all child items of this parent item.
- **Global asset**
  Only one global asset is created for all items of the same type. Therefore the permissions set for one item also affects all other items.
  (Example: Permissions set for the Tag with ID=2 will automatically be set also for the Tag with ID=3, 4, 5... Therfore all Tags will have the same set of permissions.)

This PR should also fix issue #285.

### How to test this PR
Create, edit and delete items for all possible types: `category, collection, comment, config, image, imagetype, tag, user, vote`

Collections, users votes and comments currently can only be created when migrating these items from JG3.
Therfore I reccommend to test each asset creation with one possible content type and then additionally perform a migration.

Test these three content types to test each asset creation
- **Default assets**: `Configuration sets`
- **Parent handled assets**: `Images`
- **Global asset**: `Tags` or `Imagetypes` within the configuration set


rowi68:
PR lässt sich (beim ersten Mal) nicht installieren. Fehlermeldung:

 Failed opening required '/...xxx/administrator/components/com_joomgallery/src/Table/MultipleAssetsTableTrait.php' (include_path='.:/usr/lib/php8.2') 
2024-12-27_07h51_41

Wenn ich noch einmal drüberinstalliere, funktioniert es jedoch... 🤔

eumel1602:
Das kann ich genau so (wie bei rowi) bestätigen. J4.9.x und J5.2. mit php 8.1 und php 8.2
Beim jeweils "drüberinstallieren" geht der install..
Allerdings funktioniert dann wohl nix. Neue bilder, Kategorie, alles geht nicht! 

`Es ist ein Fehler aufgetreten.

    0 Failed opening required '/var/www/web739/html/j4testversion/administrator/components/com_joomgallery/src/Table/MultipleAssetsTableTrait.php' (include_path='.:/opt/php/8.1.19/share/pear') `

MrMusic:
>This PR should also fix issue https://github.com/JoomGalleryfriends/JG4-dev/issues/285.

The following error occurs now when deleting a custom image type:
```
1 Action not available for this asset. Access can not be checked. Please provide reasonable inputs. 

Call stack
# 	Function 	Location
1 	() 	JROOT\administrator\components\com_joomgallery\src\Service\Access\Access.php:191
2 	Joomgallery\Component\Joomgallery\Administrator\Service\Access\Access->checkACL() 	JROOT\administrator\components\com_joomgallery\src\Model\JoomAdminModel.php:663
3 	Joomgallery\Component\Joomgallery\Administrator\Model\JoomAdminModel->canDelete() 	JROOT\libraries\src\MVC\Model\AdminModel.php:833
4 	Joomla\CMS\MVC\Model\AdminModel->delete() 	JROOT\administrator\components\com_joomgallery\src\Model\ConfigModel.php:424
5 	Joomgallery\Component\Joomgallery\Administrator\Model\ConfigModel->save() 	JROOT\libraries\src\MVC\Controller\FormController.php:676 
```
Migration:
When starting migration (switch to step 2) - Check migration capability, the following message always appears:
`Step 2: Migration pre-check failed. Failed check: Image mapping`
although the mapping is set correctly.

>Global asset: Tags

I think **Tags** should belong to 'Default assets' rather than 'Global asset'.
Since each tag contains the 'Permissions' tab, this would otherwise lead to confusion.

Collections: 
This is a new content type that works similar to 'Favorites' in JG3. I think it would be helpful for everyone if you could explain in more detail what can be done with them and how they are managed in the future. Thanks.

administrator/com_joomgallery/src/Service/Access/Access.php
```diff
@@ -70,6 +70,13 @@ class Access implements AccessInterface
    */
   protected $parent_dependent_types = array('image');
 
+  /**
+   * List of content types wich have only one global asset
```
brianteeman, line 74:
```suggestion
   * List of content types which have only one global asset
```

administrator/com_joomgallery/src/Service/Access/Base/AccessOwn.php
```diff
@@ -32,6 +32,14 @@ class AccessOwn extends Access
    */
   public static $parent_dependent_types = array();
 
+  /**
+   * ist of content types wich have only one global asset
```
brianteeman, line 36:
```suggestion
   * List of content types which have only one global asset
```

Elfangor93:
> I think Tags should belong to 'Default assets' rather than 'Global asset'.
Since each tag contains the 'Permissions' tab, this would otherwise lead to confusion.

No. I am heavily against that. I expect a lot of tags being created for large image galleries. But then the same performance problem will arise with Tags as we have seen happening with images. See #252.
What we could do is adding an info box at the tag edit view to inform the user that the same permission tab is displayed for all tags.

Elfangor93:
> This is a new content type that works similar to 'Favorites' in JG3.

Yes. Its just more flexible as you can have multiple 'collections' where you can add images to. The same way you will be able to add and remove images to the 'Favorites' collection (similarly as in JG3) you can have different collections where you can add images to. This way you can create your own custom galeries or image collections with your favorite images from all over the JoomGallery images.
Does this help you?

Elfangor93:
> The following error occurs now when deleting a custom image type

Should be fixed now.

> When starting migration (switch to step 2) - Check migration capability, the following message always appears:
Step 2: Migration pre-check failed. Failed check: Image mapping
although the mapping is set correctly.

I am not able to reproduce this issue. Can you provide me more details:
- Which imagetypes to you have?
- Which migration setting to you have?
- Provide a printscreen of the image mapping.

MrMusic:
>No. I am heavily against that. I expect a lot of tags being created for large image galleries. But then the same performance problem will arise with Tags as we have seen happening with images. See #252.

I can't imagine anyone creating >10,000 tags, but I don't care.

>What we could do is adding an info box at the tag edit view to inform the user that the same permission tab is displayed for all tags.

An info box is definitely necessary.


>The following error occurs now when deleting a custom image type
Should be fixed now.

Error is fixed. Thanks.

>I am not able to reproduce this issue. Can you provide me more details:

This error also no longer occurs.

What surprises me:
The table #_joomgallery_collections does not contain the column 'asset_id'. Shouldn't this be necessary for the 'Default assets' type?

It would probably make sense to merge this PR now and carry out further tests with PR https://github.com/JoomGalleryfriends/JG4-dev/pull/317

Issue #341, opened 2025-01-13
[Edit Image] Inconsistency between DB and Filesystem in case of a filesystem error


Elfangor93:
When editing and saving an image, in case of...
- Category was changed
- Image gets copied
- Alias has changed
... and there is an error in the filesystem trying to adjust the files and folders to adapt to this changes, then the Filestystem creates an error, but the database is already changed. This leads to incosistency between database and filesystem.

### What to do to fix this issue
Add a possibility to roll back the database in case there is an error in the filesystem after the database has already changed.

Elfangor93:
The same issue also is present for categories. 🙈

PR #343, opened 2025-01-14
[Image save] Change files in filesystem before store to DB

Elfangor93:
This PR applies one possible solution to solve the issue reported in #341.

All changes in the file and folders are processed before the item gets stored to the database. If an error appears while changing things in the filesystem, the item will not be stored and the save process is cancelled.

### How to test this PR
Create, Edit, Delete, Move, Rename Images and Categories.
Everything should work as before.
Optional: Check that Database and Folderstructure are consistent.


MrMusic:
I noticed the following problem during testing: Moving images to another category is not possible if no original image file is available.
1. switching off creating 'original' images in the JG configuration
2. uploading images
3. attempts to move an image to another category
4. error message: “Save failed with the following error: ”
5. But detail image and thumbnail have been moved to the other folder

MrMusic:
Unable to change an image alias.
If you try to change the alias of a saved image, the following error message appears: 
`Save failed with the following error: `
For categories, the alias can be changed without any problems.

Elfangor93:
> Unable to change an image alias.

I can not reproduce this issue. Changing the alias is perfectly possible.

MrMusic:
>I can not reproduce this issue. Changing the alias is perfectly possible.

It depends on what is set in the JoomGallery configuration for use 'Original filename'.

1. Set 'Original filename' to Yes in the JG configuration
2. Uploading images
3. Attempts to change an image alias
    Error message: "Save failed with the following error: "
    ->New alias cannot be saved

Elfangor93:
> Unable to change an image alias.

@MrMusic Is fixed now.

MrMusic:
>@MrMusic Is fixed now.

Thanks.

Thank you very much!

Issue #342, opened 2025-01-13
[Filesystem] Not possible to move or copy images between different filesystems


Elfangor93:
There is no possibility currently to move or copy images between different filesystems. Once uploaded the image stays in the filesystem where it was created.

The two possibilities to work around this issue:
1. Delete the image and reuplod it on another filesystem
2. Edit the image form corespondingly. This will create a filesystem error, but the DB was correctly changed. Then adjust the filesystem by hand.

### How to solve this issue
Expand the Filesystem Service to handle two different filesystem adapters and use the correct one depending on the task to perform.

PR #351, opened 2025-01-25
Fix: Move image to another filesystem

Elfangor93:
This PR adds an error handling for the issue #342.

If you try to move or copy an image into a category with another filesystem, you will get an error and the save process is cancelled. This is not a proper fix, but it prevents corrupt datasets. For a proper fix, the filesystem service would have to be extended to support multiple filesystems. This can be implemented in a later release...

### How to test this PR
- When editing an existing category, the parameter field `jg_filesystem` has to be deactivated
- When changing the category of an existing image to a category having another filesystem set in its configuration, you will get an error and the image will not be saved.
- When copying a existing image and select a category having another filesystem set in its configuration, you will get an error and the image will not be saved.


MrMusic:
Thank you very much!

PR #357, opened 2025-01-30
Default pagination for subcategories

reilldesign:
This pull request brings pagination for the standard display. Load more and infinite scroll will follow in a later pull request.

this addresses issue #347

### DIFF
--- a/site/com_joomgallery/src/View/Category/HtmlView.php
+++ b/site/com_joomgallery/src/View/Category/HtmlView.php
@@ -98,8 +98,6 @@ class HtmlView extends JoomGalleryView
 			$this->item->children = new \stdClass();
 			$this->item->children->items         = $this->get('Children');
 			$this->item->children->pagination    = $this->get('ChildrenPagination');
-			$this->item->children->filterForm    = $this->get('ImagesFilterForm');
-			$this->item->children->activeFilters = $this->get('ImagesActiveFilters');
 
 			// Load images
 			$this->item->images = new \stdClass();
--- a/site/com_joomgallery/src/View/Category/HtmlView.php
+++ b/site/com_joomgallery/src/View/Category/HtmlView.php
@@ -96,8 +96,11 @@ class HtmlView extends JoomGalleryView
 
 			// Load subcategories
 			$this->item->children = new \stdClass();
-			$this->item->children->items = $this->get('Children');
-	
+			$this->item->children->items         = $this->get('Children');
+			$this->item->children->pagination    = $this->get('ChildrenPagination');
+			$this->item->children->filterForm    = $this->get('ImagesFilterForm');
+			$this->item->children->activeFilters = $this->get('ImagesActiveFilters');
+
 			// Load images
 			$this->item->images = new \stdClass();
 			$this->item->images->items         = $this->get('Images');
--- a/administrator/com_joomgallery/forms/config.xml
+++ b/administrator/com_joomgallery/forms/config.xml
@@ -114,7 +114,7 @@
           <option value="0">JNO</option>
           <option value="1">JYES</option>
       </field>
-      
+
       <field name="jg_router"
              type="jgdynamiclist"
              default="DefaultRouter"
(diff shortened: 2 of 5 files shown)


eumel1602:
Alwin's error description in #347 is no longer there.
It works.
Switching the pagination works.
Couldn't find any errors in this PR

site/com_joomgallery/src/View/Category/HtmlView.php
```diff
@@ -96,8 +96,11 @@ public function display($tpl = null)
 
 			// Load subcategories
 			$this->item->children = new \stdClass();
-			$this->item->children->items = $this->get('Children');
-	
+			$this->item->children->items         = $this->get('Children');
+			$this->item->children->pagination    = $this->get('ChildrenPagination');
+			$this->item->children->filterForm    = $this->get('ImagesFilterForm');
```
Elfangor93, line 101: Wozu brauchst du das Filter-Formular?
Ich sehe keine Verwendung im Template...

site/com_joomgallery/src/View/Category/HtmlView.php
```diff
@@ -96,8 +96,11 @@ public function display($tpl = null)
...
+			$this->item->children->activeFilters = $this->get('ImagesActiveFilters');
```
Elfangor93, line 102: Brauchst du das Filter-Formular?
Ich sehe keine Verwendung im Template...

Elfangor93 (dismissed):

site/com_joomgallery/src/View/Category/HtmlView.php
reilldesign: Die Filter werden aktuell nicht benötigt und ich habe sie entfernt.

MrMusic:
Thank you very much!

Elfangor93:
The critical test is to visit a category having a subcategory pagination and an images pagination at the same page. Check there that the two paginations are not interfering with each other.

MrMusic:
>Check there that the two paginations are not interfering with each other.

Yes, you are right, the two paginations interfere with each other. The problem occurs when there are a large number of subcategories and a large number of images:
- if I turn to the next page for the subcategories, the pagination of the images starts again on the first page
- if I turn to the next page for the images, the pagination of the subcategories starts again on the first page
- if I set something other than 'Joomla default pagination' for the images, the pagination for the subcategories disappears

Elfangor93:
> if I turn to the next page for the subcategories, the pagination of the images starts again on the first page
    if I turn to the next page for the images, the pagination of the subcategories starts again on the first page
    if I set something other than 'Joomla default pagination' for the images, the pagination for the subcategories disappears


Fixing this is not trivial. Do we wanna take the time to fix this and release the stable version later or do we wanna keep this bug and fix it in the v4.1 release?

reilldesign:
The subcategories navigation types Load More and Infinite scroll should only be added in version 4.1.

MrMusic:
At the moment it looks like this:

subcategories pagination | images pagination | works
-- | -- | --
Joomla default pagination | Joomla default pagination | Ok
Joomla default pagination | Infinity scroll | ?
Joomla default pagination | Load more button | ?
Infinity scroll | Joomla default pagination | in 4.1
Infinity scroll | Infinity scroll | in 4.1
Infinity scroll | Load more button | in 4.1
Load more button | Joomla default pagination | in 4.1
Load more button | Infinity scroll | in 4.1
Load more button | Load more button | in 4.1

The combinations 
Joomla default pagination | Infinity scroll
Joomla default pagination | Load more button
should still come with the release of 4.0 ?

Elfangor93:
@reilldesign I can not really help here as I still dont fully understand how `Infinity scroll` and `Load more` are working. Therefore I can not judge how much work it is to make these combinations work.

reilldesign:
@Elfangor93 @MrMusic there was still an error in the navigation query for the subcategories. Now all three category navigation types work with the standard navigation for the subcategory.

MrMusic:
@reilldesign  Thank you very much, it works very well now!
One last request: I would recommend to set the 'Pagination' field in the configuration to 'Readonly' to prevent a different/incorrect selection.

reilldesign:
@MrMusic the setting option for subcategory pagination is set to readonly true up to version 4.1.

MrMusic:
Thank you very much!

Elfangor93:
@reilldesign Kann ich diesen PR so mergen?

reilldesign:
@Elfangor93 sehr gerne mergen.

Issue #356, opened 2025-01-30
Error when using real path


hansdampf-123:
Beta 2:
If Common Settings - Use real path is set to YES
the categoriey dont open anymore, but this error is shown:

Joomgallery\Component\Joomgallery\Administrator\Service\Filesystem\Filesystem::__construct(): Argument #1 ($filesystem) must be of type string, null given, called in XXX/administrator/components/com_joomgallery/src/Service/Filesystem/FilesystemServiceTrait.php on line 56

Elfangor93:
Can you please post the whole call stack of this error?

See: https://www.joomgalleryfriends.net/en/documentation/finding-errors-en/isolate-joomgallery-problems.html

hansdampf-123:
Joomla 5.2.3
PHP 8.1.31

<head></head>
# | Function | Location
-- | -- | --
1 | () | JROOT/administrator/components/com_joomgallery/src/Service/Filesystem/Filesystem.php:83
2 | Joomgallery\Component\Joomgallery\Administrator\Service\Filesystem\Filesystem->__construct() | JROOT/administrator/components/com_joomgallery/src/Service/Filesystem/FilesystemServiceTrait.php:56
3 | Joomgallery\Component\Joomgallery\Administrator\Extension\JoomgalleryComponent->createFilesystem() | JROOT/administrator/components/com_joomgallery/src/Helper/JoomHelper.php:105
4 | Joomgallery\Component\Joomgallery\Administrator\Helper\JoomHelper::getService() | JROOT/administrator/components/com_joomgallery/src/Helper/JoomHelper.php:528
5 | Joomgallery\Component\Joomgallery\Administrator\Helper\JoomHelper::getImg() | JROOT/components/com_joomgallery/layouts/joomgallery/grids/images.php:53
6 | include() | JROOT/libraries/src/Layout/FileLayout.php:128
7 | Joomla\CMS\Layout\FileLayout->render() | JROOT/libraries/src/Layout/LayoutHelper.php:76
8 | Joomla\CMS\Layout\LayoutHelper::render() | JROOT/components/com_joomgallery/tmpl/category/default_cat.php:245
9 | include() | JROOT/libraries/src/MVC/View/HtmlView.php:416
10 | Joomla\CMS\MVC\View\HtmlView->loadTemplate() | JROOT/components/com_joomgallery/tmpl/category/default.php:33
11 | include() | JROOT/libraries/src/MVC/View/HtmlView.php:416
[...]

PR #358, opened 2025-02-01
Fix two frontend PHP errors

Elfangor93:
This PR fixes the fowllowing two frontend errors:
- Detail view: PHP error when having an array in some metadata values
- Category & Gallery view: PHP error when activating `Use real paths` in component configurations
(#356)

### How to test this PR
- [This image](https://drive.google.com/file/d/1hnef5zEADPxv7GFAuILAxLnbEOsr9Zwu/view?usp=sharing) creates an error in the single image view when this PR is not applied. No error with this PR.
- Activate `Use real paths` in component configurations and visit a category view containing at least one image.


MrMusic:
Thank you very much!

PR #362, opened 2025-02-01
Fix wrong default value

MrMusic:
After a new installation, two columns in the configuration ('jg_gallery_view_image_link' and 'jg_category_view_image_link') have incorrect default values.
Therefore, the images in the 'Gallery view' and in the 'Category view' have no link after the installation. The link only works after changing the configuration.

### How to test
Perform a new installation
Load images into a category
Check in the frontend: 

### Before this PR
The images in the 'Gallery view' and in the 'Category view' have no link.

### After this PR
The images in the 'Gallery view' and in the 'Category view' has now link to the 'Single image view'.


eumel1602:
Ich habe diesen PR erfolgreich getestet 
DANKE

PR #373, opened 2025-02-16
Fix issue with xml path

Elfangor93:
This issue fixes an error reported by a JoomGallery user reported us that the configuration set edit view can not be opened when the language debugging is active.

grafik

### How to test this PR
After applying this PR you can now open the "Global Configuration" without error even if Joomla & Language debugging is activated.


eumel1602:
ich habe den Pr "drüber" installiert und auch komplett neu in Joomla 5.2.3. (php 8.2.5)
die Config. geht **NICHT zu öffnen**:

Fehler call strack:

Elfangor93:
@eumel1602 Is fixed now.

BennoMath:

eumel1602:
I have now successfully tested the PR ✅

PR #379, opened 2025-02-22
Fix issue with asset_id during migration

Elfangor93:
This PR fixes the issue reported in issue #378.

It seems that there was a change in Joomla 5.2.4 onwards wich requires to have an integer value as the `asset_id` when creating a new asset entry. Therefore now we introduce `asset_id = 0` instead of `asset_id = null`.


MrMusic:
Thank you very much!

